PR #2, opened 2015-01-11
created first R problem and scripts to automate testing with RUnit

morphatic:
Created a basic "Hello, World!" exercise for R.

Also created some scripts that will automate the testing process so that students will not have to understand RUnit to get started. It should work for both Linux/OSX and Windows.  I tested it on OSX and Windows.

I also wrote [instructions for getting started with R](https://github.com/morphatic/docs/blob/master/app/pages/languages/getting-started-with-r.md) that can be merged into the [docs project](https://github.com/exercism/docs).


kytrinyx:
We're going to need files for `hello-world` in the github.com/exercism/x-common repository, otherwise the x-api will blow up when I add this.

morphatic:
Right.  Done. https://github.com/exercism/x-common/pull/43

kytrinyx:
Would you mind rebasing this onto master? I've gotten all the basic infrastructure worked out for a "hello world" problem across the board, including an automated bit of feedback from rikki: https://github.com/exercism/rikki/blob/master/comments/hello/hello.md

kytrinyx:
@morphatic sorry it took me so long to get this sorted.

Would you mind splitting this up into several PRs?

I can merge the hello-world exercise (since that has the requisite files in the x-common repo).

If you would add the metadata for `hello-variables` and `operate` to `x-common` then I could merge those exercises as well.

config.json
```diff
@@ -1,18 +1,17 @@
 {
   "slug": "r",
   "language": "R",
-  "repository": "https://github.com/exercism/xr",
-  "active": false,
+  "repository": "https://github.com/morphatic/xr",
+  "active": true,
```
kytrinyx: Please don't activate the track until there are at least 10 exercises.

config.json
```diff
@@ -1,18 +1,17 @@
 {
   "slug": "r",
   "language": "R",
-  "repository": "https://github.com/exercism/xr",
-  "active": false,
+  "repository": "https://github.com/morphatic/xr",
```
kytrinyx: The repository needs to be `exercism/xr`, as that is the canonical repo that the x-api uses to serve problems from.

kytrinyx:
One more thing -- would you mind separating the hello-world commit from the track-level CI script for running the tests, please? These are not related to each other.

kytrinyx:
In case you're interested, here's my favorite resource about commit messages: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

It assumes that each commit makes a single, atomic, cohesive change.

That makes it much, much easier to manage the project.

In terms of these language tracks, I tend to go for commits that look something like this:

```
implement hello-world exercise
```

Or:

```
hello-world: fix inconsistency in test names
```

This makes it easy to go through the git history later.

Issue #3, opened 2015-03-16
Add helpful information to the SETUP.md


kytrinyx:
The contents of the SETUP.md file gets included in
the README.md that gets delivered when a user runs the `exercism fetch`
command from their terminal.

At the very minimum, it should contain a link to the relevant
language-specific documentation on
[help.exercism.io](http://help.exercism.io).

It would also be useful to explain in a generic way how to run the tests.
Remember that this file will be included with _all_ the problems, so it gets
confusing if we refer to specific problems or files.

Some languages have very particular needs in terms of the solution: nested
directories, specific files, etc. If this is the case here, then it would be
useful to explain what is expected.

------

Thanks, @tejasbubane for suggesting that we add this documentation everywhere.
See [exercism.io#2198](https://github.com/exercism/exercism.io/issues/2198).

rebelwarrior:
The setup for R doesn't work. There is no run_tests utilities.

kytrinyx:
@jonboiser would you have time to look at this?

rebelwarrior:
I did some digging and the issue is that run_utilities is a separate file that's included with hello world but not word-count. This seems like a complicated setup. And the information on how to set it up is on the hello world readme. That information (with more detail) should reside in the setup file.

rebelwarrior:
I found the error and submitted a pull request. It was on TEST.md. There was a reference to "run_tests" that seems to be a different way of running the tests. The way it should say is `Rscript <testfile>`

jonboiser:
@rebelwarrior Thanks for pointing this out. I forgot to update TEST.md when I was simplifying the problem folders to just have a few files.

I updated the TEST.md with the updated info and did a scan through all the docs to see if there was any more contradicting info.

PR #8, opened 2015-12-22
Setup and installation instructions for R

jonboiser:
Relates to #3. I'll squash and rebase this when I'm done.

I'd like to pick up the torch from the earlier contributor. These (WIP) setup instructions describe a pretty mainstream path to setting up an R environment based around RStudio. I think this is a good way to go, as there is a lot of documentation and e-learning resources based around this IDE. 

One kind of significant departure from the original PR is that I'd like to use the `testthat` test runner instead of `runit`, as this seems to be more popular for R library writers and has some good docs and RStudio support.

I plan on writing some instructions on running tests through the command-line too.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,41 @@
+## Installation
+See [this guide](https://github.com/exercism/xr/blob/master/docs/INSTALLATION.md) for instructions on how to setup your local R environment.
+
+## How the R files are organized and why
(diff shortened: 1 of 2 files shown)


kytrinyx:
This looks great. I've given you commit to the repo. If you want me to give anyone else commit as well, let me know.

PR #12, opened 2016-05-09
Make word-count expectations tolerant to differences in the order of the word counts

Robsteranium:
When `expect_equal` compares the test object and an expected value it requires that the order of the named elements in those lists be the same.

That requirement is not necessary for the `word-count` function. This commit makes the tests less specific to this implementation detail.


kytrinyx:
Good catch, thank you.

PR #22, opened 2016-12-02
add difference-of-squares exercise

cmiller01:


kytrinyx:
Thank you @cmiller01!

PR #23, opened 2016-12-02
Add hamming

cmiller01:
i included difference-of-squares in this PR as well, just turned out to be easier and figured the config.json file might conflict if these were separate. happy to break up if that's better.

### DIFF
--- a/config.json
+++ b/config.json
@@ -44,6 +44,11 @@
       "difficulty": 1,
       "slug": "word-count",
       "topics": []
+    },
+    {
+      "difficulty": 1,
+      "slug": "difference-of-squares",
+      "topics": []
     }
   ]
 }
--- a/config.json
+++ b/config.json
@@ -49,6 +49,11 @@
       "difficulty": 1,
       "slug": "difference-of-squares",
       "topics": []
+    },
+    {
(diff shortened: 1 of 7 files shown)


cmiller01:
@jonboiser @kytrinyx wondering if you could take a look at this? Would like to get more exercises into the R track.

kytrinyx:
Thanks @cmiller01—I'm looking at these today. I've merged difference of squares and will pull this down locally so we just get hamming in this PR.

I think that `config.json` won't conflict if you put some things earlier in the list (as opposed to at the end). This does affect the order that the exercises are delivered, so sometimes it won't make sense to put things earlier, but if it does make sense that would be a way to avoid the conflicts.

kytrinyx:
Merged in d4c904de1abec5811227fca0cd97d3438c378d6b

cmiller01:
Thanks @kytrinyx, good call on the ordering exercises in config.json. I'll work on getting more of these created. Would love for R not to be an orphaned repo for exercism.

kytrinyx:
> Would love for R not to be an orphaned repo for exercism.

I would love that too! :heart:

PR #25, opened 2016-12-18
add working examples, update stub for luhn

cmiller01:
adds examples for exercises, which were missing
* anagram
* bob
* word-count
* luhn


cmiller01:
@jonboiser do you think this is good to go? want to update the travis build script and would like to get this in first...

PR #26, opened 2016-12-19
fix test cases + add example for space-age

cmiller01:
the test cases for space-age isn't correct, updating to take planet as argument. adding example as well


jonboiser:
thanks for the PR. i'll take a look at the luhn one shortly

cmiller01:
no problem, thanks for the review and the formatting fixes!

PR #28, opened 2017-01-31
Add leap exercise

jonmcalder:


jonboiser:
Thanks for the PR! Would you mind adding entries in config.json for the other two exercises, plus added a file-ending newline to all of the new scripts.

Once everything is squared away for these three exercises, I'll merge them in. Be glad if you have more exercises you want to contribute to get the R path closer to a releasable state.

jonmcalder:
Sure no problem - will remember in future.

Should be able to do a few more exercises during the week. Would be nice if we can get the R path released sometime soon!

Issue #33, opened 2017-02-19
docs/SETUP.md: should be moved to exercises/TRACK_HINTS.md


petertseng:
For the purpose of #10, `SETUP.md` in the root of this repo was moved to `docs/SETUP.md`.

Exercism does not assign special meaning to `docs/SETUP.md` (it **will not** be shown anywhere on http://exercism.io/languages/r) but it does assign special meaning to `SETUP.md` - it gets included in the README for each problem delivered to the student.

In addition, because of https://github.com/exercism/exercism.io/issues/3240, `SETUP.md` should get moved to `exercises/TRACK_HINTS.md`.

So, the two steps

1. allowing the instructions to be shown to the student again (`docs/SETUP.md` -> `SETUP.md`)
2. moving to the new location (`SETUP.md` -> `exercises/TRACK_HINTS.md`)

the file should be moved to `exercises/TRACK_HINTS.md`

jonmcalder:
Thanks!

PR #34, opened 2017-02-20
Move '/docs/SETUP.md' to '/SETUP.md' and '/exercises/TRACK_HINTS.md' …

jonmcalder:
…- this closes issue #33

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,5 @@
+## Installation
+See [this guide](https://github.com/exercism/xr/blob/master/docs/INSTALLATION.md) for instructions on how to setup your local R environment.
+
+## How to implement your solution
+In each problem folder, there is a file named `<exercise_name>.R` containing a function that returns a `NULL` value. Place your implementation inside the body of the function.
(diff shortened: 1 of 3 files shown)


jonmcalder:
I have split these out based on the instructions in #33 and on a cursory look at the xr/go repo which references the testing procedure in the exercises/TRACK_HINTS.md file. 

We may want to review the content in each of these two files to make sure it is appropriate given where and how it will appear on the site.

So far I have just replaced "problem.R" and "test_problem.R" with "<exercise_name>.R" and "test_<exercise_name>.R" to highlight that these file names change for each problem.

petertseng:
Note that if this track (any track) has **both** `SETUP.md` and `exercises/TRACK_HINTS.md`, by https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L105 **only `exercises/TRACK_HINTS.md` will be used**. Since the latter is meant to replace the former, it's one or the other.

jonmcalder:
Oh ok - I had interpreted your instructions to mean add both not one or the other. Will move everything into exercises/TRACK_HINTS.md then.

Thanks!

PR #39, opened 2017-03-30
Added exercise

llenk:
Added prime factors exercise


llenk:
I fixed those issues. Is there anything else I should change?

PR #46, opened 2017-05-21
cosmetic update

katrinleinweber:


jonmcalder:
Thanks for this @katrinleinweber - I think there is a lot of code in this repo that doesn't yet follow/adhere to the suggested style guide. I'm hope to review and correct this at some stage, but help like this is appreciated as it'll mean less for me to do when I get around to it! :-)

PR #47, opened 2017-05-22
more cosmetics: apply RStudio's formatter & lintr

katrinleinweber:
follow-up to #46

- [x] check for inadvertent replacement of `'` within natural language string
- [x] address all `lint()` markers
There are still warnings, though: `In seq.default(x, b = ncol(res), length.out = 3) :  partial argument match of 'b' to 'by'`. I guess they are due to some code called by `lintr`. Found similar report at https://github.com/ggrothendieck/sqldf/issues/8, but not sure whether & how to investigate further.


jonmcalder:
Thanks this looks good! Haven't got time to review this now but will follow up on it later.

jonmcalder:
I can't seem to reproduce those warnings. Can you clarify further what you did?

katrinleinweber:
The warnings appeared when I `setwd("./exercises)` and used `list.files(".", pattern = "R$", recursive = TRUE) %>% lapply(lintr::lint)`.

jonmcalder:
Ok - I'm still not seeing those warnings in my environment, so I'm gonna assume for now that it's just some kind of oddity.

katrinleinweber:
OK, so much for the tests ;-)   What about the examples? Since they seem a bit more personal, I'm not sure whether linting/reformatting them is appropriate. If yes, what would be the best way in your opinion to proceed?

- Using [@yihui's formatR](https://github.com/yihui/formatR) with default settings on everything will probably result in a lot of changes about which different exercisers have different opinions.
- RStudio's own reformatter seems to be less aggressive, but also less automatable across many files. Or am I overlooking a recursive option there?
- PrettyR? Never used it before.

Recommendations welcome, then I'll dive in. But I also understand if making-over the tests is enough for now.

jonmcalder:
Thanks for following up! Yeah I hear you - style guides are definitely a subjective thing but for open source repo's they are certainly helpful and this means some people may need to compromise on their personal preference in the interest of the collective.

It's still early days for the R track so I think there is much room for improvement of the example solutions (and not only with regards to coding style) - I haven't been through all of them but I'm sure many of the example solutions I contributed could be improved upon :-)

Given that I expect much of this code to be updated in future, my sense is that formatR is probably the best starting point for now even if it may be a little aggressive or glitchy in some regards - e.g. the known limitations outlined [here](https://yihui.name/formatr/). Once we've done that we can then resort to a more nuanced manual approach to the finer details as necessary.

There is also [styler](https://github.com/krlmlr/styler) which is a different (non-invasive) approach - a friend of mine is working on this project for Google Summer of Code this year, so I'm planning to get some input from him.

You'll see from #49 that my plan/goal is to add automated linting/testing to this repo (I intend to make use of [lintr](https://github.com/jimhester/lintr), probably with defaults which are based on @hadley's style guide).

This should then facilitate adherence to the style guide for any updates we make and for new pull requests coming in.

PR #50, opened 2017-05-25
add same parameters as in tests

katrinleinweber:
Assuming that a novice will look at the test file to determine what their code is supposed to do, the stub might as well include the same parameters.

~~Exception: new, generic parameter name in word-count.R~~


katrinleinweber:
Oh. Sorry!

jonmcalder:
No worries!

PR #57, opened 2017-06-21
Update tests from x-common

katrinleinweber:
* several tests, descriptions and examples updated from their respective `canonical-data.json` files (see commits for details)

* In order to highlight alternative code patterns, `dplyr` was tested in the `perfect numbers` example. Conclusion: stick to base R in the future and maybe comment on useful packages.

@jonmcalde: I'm fine with keeping this open until all tests suites are reviewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
 - R -e "install.packages('jsonlite')"
 - R -e "install.packages('lintr')"
 - R -e "install.packages('magrittr')"
+- R -e "install.packages('dplyr')"
 script:
 - bin/fetch-configlet
 - bin/configlet .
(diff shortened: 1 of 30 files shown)


lorenzwalthert:
Because `return()` will return an expression and not continue with evaluating the remainder of the code

``` r
f <- function() {
  return(1)
  stop("error")
}
f()
#> [1] 1
```

According to the [tidyverse style guide](http://style.tidyverse.org/functions.html#return), you should use it for early termination only. I think using it within other functions should best be avoided since hardly necessary.

katrinleinweber:
Agreed, will do! But something in the script or my environment _must_ have changed that behaviour. Otherwise, I would have seen the same errors as Travis.

PR #59, opened 2017-06-23
Add pangram, nucleotide-count, triangle & collatz-conjecture

jonmcalder:
Any thoughts on this approach to the triangle exercise? I've gone with a simple S3 implementation, but it could instead be done as just a simple function which returns the type of the triangle as a string.

See [this discussion](https://github.com/exercism/problem-specifications/issues/379) for more details and context.

### DIFF
--- a/exercises/collatz-conjecture/test_collatz-conjecture.R
+++ b/exercises/collatz-conjecture/test_collatz-conjecture.R
@@ -1,0 +1,28 @@
+source("./collatz-conjecture.R")
+library(testthat)
+
+test_that("zero steps for one", {
+  expect_equal(collatz_step_counter(1), 0)
+})
+
+test_that("divide if even", {
+  expect_equal(collatz_step_counter(16), 4)
+})
+
+test_that("even and odd steps", {
+  expect_equal(collatz_step_counter(12), 9)
+})
+
+test_that("Large number of even and odd steps", {
+  expect_equal(collatz_step_counter(1000000), 152)
+})
+
+test_that("zero is an error", {
+  expect_error(collatz_step_counter(0))
+})
+
+test_that("negative value is an error", {
+  expect_error(collatz_step_counter(-15))
+})
+
+message("All tests passed for exercise: collatz-conjecture")
--- a/exercises/collatz-conjecture/test_collatz-conjecture.R
+++ b/exercises/collatz-conjecture/test_collatz-conjecture.R
@@ -1,27 +1,27 @@
 source("./collatz-conjecture.R")
 library(testthat)
 
-test_that("zero steps for one", {
+test_that("Input of one results in zero steps", {
   expect_equal(collatz_step_counter(1), 0)
 })
 
-test_that("divide if even", {
+test_that("Even input with repeated even steps", {
   expect_equal(collatz_step_counter(16), 4)
 })
 
-test_that("even and odd steps", {
+test_that("Input which results in small number of even and odd steps", {
   expect_equal(collatz_step_counter(12), 9)
 })
 
-test_that("Large number of even and odd steps", {
+test_that("Input which results in large number of even and odd steps", {
   expect_equal(collatz_step_counter(1000000), 152)
 })
 
-test_that("zero is an error", {
+test_that("Input of zero results in an error", {
   expect_error(collatz_step_counter(0))
 })
 
-test_that("negative value is an error", {
+test_that("Negative input results in an error", {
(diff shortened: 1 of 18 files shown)


katrinleinweber:
Hi Jon! I think I'll need well into July to implement the exercises myself. Right now, I'm indifferent towards the triangle implementation.

About reviewing: I skimmed through https://github.com/exercism/docs/blob/master/maintaining-a-track/reviewing-a-pull-request.md#reviewing-a-pull-request, but feel that it is geared more towards PRs in the general `problem-specifications`.

exercises/nucleotide-count/example.R
```diff
@@ -0,0 +1,17 @@
+library(magrittr)
```
katrinleinweber: I like this pkg very much and using the  ["standard library" is only mentioned with regards to the tests](https://github.com/exercism/docs/blob/master/maintaining-a-track/reviewing-a-pull-request.md#encourage-best-practices). I think because piping is becoming more popular in the R community, we should start including it in examples. Maybe not re-write old examples, though ;-)

katrinleinweber:
I have an uneasy feeling about the test case descriptions for [`nucleotide-count`](https://github.com/exercism/problem-specifications/blob/master/exercises/nucleotide-count/canonical-data.json) and [`collatz-conjecture`](https://github.com/exercism/problem-specifications/blob/master/exercises/collatz-conjecture/canonical-data.json). Many do not tell me clearly, what the expectation or failure exactly is. However, because I haven't implemented against the tests yet, I'm not sure if how they could be improved (preferably upstream).

Still, because the guideline summary is ["[…] we'll see [and] improved later."](https://github.com/exercism/docs/blob/master/maintaining-a-track/reviewing-a-pull-request.md#reviewing-a-pull-request), I opt for merging (without squash or rebase).

exercises/pangram/example.R
```diff
@@ -0,0 +1,14 @@
+library(magrittr)
+
+is_pangram <- function(input) {
+
+  input_letters <- gsub("[^A-Za-z]", "", input) %>% 
+                    tolower() %>% 
+                    strsplit("") %>% 
+                    unlist() %>% 
+                    unique() %>% 
+                    sort()
```
katrinleinweber: One can omit empty `()` for functions that are part of a pipe, but in the interest of differentiating it them from other variables, we should keep them in the examples.

exercises/nucleotide-count/example.R
jonmcalder: Yeah I agree. Like you say the use of piping in R has become more common over time. Plus in many circles it's probably considered good practice since it improves readability. And ultimately the examples aren't really user facing anyway so it shouldn't matter too much.

exercises/pangram/example.R
jonmcalder: 👍

jonmcalder:
I actually agree with your comment on the vague test case descriptions - I didn't really give it any thought initially since I was just replicating what was present in the canonical tests, but I do think a few of them could be altered to offer a better description of what aspect of the problem is being tested and what is expected. I think I'll try and alter these slightly to improve on this.

jonmcalder:
I have made some updates and rebased onto master:
- updated descriptions for the test cases slightly to (hopefully) make them a little clearer
- also updated the config based on the new metadata requirements
  - added topics
  - for now (see below), used the defaults for track progression (core: `false` & unlocked_by: `null`)

Will keep this on hold until we are done with #69 though. Then we can decide where to incorporate these exercises into the overall track progression.

exercises/collatz-conjecture/test_collatz-conjecture.R
```diff
@@ -0,0 +1,28 @@
+source("./collatz-conjecture.R")
+library(testthat)
+
+test_that("Input of one results in zero steps", {
```
katrinleinweber: I had to read this several times, because `of one results` seemed strange. I guess it's grammatically correct, but how about using just `1` here, not the word. Same for the `of zero results` below.

jonmcalder: Haha - fair enough. To be honest I did try and re-write these a few different ways and wasn't really sure...

jonmcalder:
Thanks for the review!

exercises/collatz-conjecture/test_collatz-conjecture.R
jonmcalder: Done.

PR #60, opened 2017-06-29
add beer-song

katrinleinweber:
- [x] follow https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/porting-an-exercise.md
- [x] port tests: https://github.com/exercism/problem-specifications/blob/master/exercises/beer-song/canonical-data.json
- [x] implement an example


katrinleinweber:
This was exhausting. I went for a deduplicated example twice, before breaking out of the optimisation habit I got into while working through the earlier exercises myself m-)

@jonmcalder Ready to be reviewed :-) We could use a label for that in this repo's settings, BTW ;-)

config.json
```diff
@@ -178,6 +178,13 @@
       "topics": [
 
       ]
+    },
+    {
+      "difficulty": 2,
+      "slug": "beer-song",
```
jonmcalder: I think this problem should fall much earlier in the track. The main aspect of this is string substitution/manipulation (which is not particularly difficult) and this is paired with looping & conditionals which should have been established pretty early on. Most of the other tracks (with a few exceptions) seem to have placed this problem somewhere in the first half of the track. We've got less exercises implemented though so for now I'd suggest somewhere around the middle near `anagram`/`pangram`/`isogram`.

I'm aware that the ordering of exercises for the track as a whole still needs more attention, but again given the upcoming redesign it's probably not worth agonizing over now - I did some re-ordering in [42](https://github.com/exercism/r/pull/42) based on a comparison of other language tracks and I noticed there was a lot of variability - so for now I think let's just aim for something 'reasonably' ordered.

katrinleinweber:
About the [squashing](https://help.github.com/articles/about-merge-methods-on-github/#squashing-your-merge-commits): I haven't used it before and it's not clear to me whether GH will let me edit the commit message. In `master`, it should contain `beer-song`, right? Or it it sufficient if the merge commit contains my branch's name `add-beer-song`?

jonmcalder:
Funnily enough neither have I actually - but I think there is a confirmation step so you can edit the commit message beforehand if you need to.

katrinleinweber:
There was :-)

Issue #62, opened 2017-07-04
add (micro)benchmark to test performance


katrinleinweber:
Several exercise descriptions suggest refactoring a solution for readability, performance etc. I suggest adding a single [`microbenchmark`](https://cran.r-project.org/web/packages/microbenchmark/index.html) to 
- a) either only those exercises, or
- b) each exercise,

in order to provide a quick feedback on the performance. Thoughts, comments, etc.?

Console output for a successful test suite is only few lines, so `microbenchmark`'s results should be visually comparable across runs. I'm not sure how the alternative (automatically benchmark 2 versions of a solution at once) would be possible.

jonmcalder:
This is a cool suggestion. 

Previously I'd wondered whether it might even be possible to include some form of formal benchmark 'testing' for certain exercises where the efficiency of the implementation could or should be part of the focus. I think that is probably tricky territory though, given that runtime is system dependent and thus somewhat variable in nature (I suppose one could build in a buffer but still).

Anyway, I'm guessing that what you're suggesting here is more of a 'soft' inclusion, just to encourage people to look at and consider benchmarking while iterating on solutions? Maybe including commented out benchmarking code or something like that?

katrinleinweber:
I have too little time to work on this further and to ensure compatibility within the existing exercism environment, sorry.

jonmcalder:
No worries! I'm also too busy to look at this at the moment. I may try and revisit it at a later stage though - thanks for the suggestion and the progress made on it so far!

PR #63, opened 2017-07-07
PoC for benchmarking

katrinleinweber:
Based on #62, here's a quick proof-of-concept to get one benchmark result into the console output, while running `testthat::auto_test()`. Strangely, `message()` doesn't produce any output, `cat()` does so twice. I'm open to improvement suggestion for this code here, or further general discussion about benchmarking in #62.


exercises/grains/test_grains.R
```diff
@@ -45,4 +45,15 @@ test_that("returns the total number of square on the board", {
   expect_equal(total(), 18446744073709551615)
 })
 
+# assist the "optimize for speed" instruction
+test_that("benchmarking", {
+  warning(
+    capture_output(
+      print = TRUE,
+      microbenchmark:::print.microbenchmark(
```
jonmcalder: You can remove this call to `print.microbenchmark` as it shouldn't be necessary.

exercises/grains/test_grains.R
```diff
@@ -45,4 +45,15 @@ test_that("returns the total number of square on the board", {
...
+        microbenchmark::microbenchmark(
+          total()
+        ))))
+})
+
 message("All tests passed for exercise: grains")
```
jonmcalder: Doing it this way kills the "All tests passed ..." message :-(

jonmcalder: Oops - my mistake. `auto_test()` never included this output originally anyway since it's just a message and not a test.

jonmcalder: Another minor request is to re-order this so that for people running the test file interactively (and on Travis) the test success message appears before the benchmarking output.

katrinleinweber:
OK, done :-) 

`print.microbenchmark` would offer the option to reduce `signif`, which I found visually helpful. Just seeing a single digit for example, instead of `123.456`. Since this is not really about benchmarking, but rather basic gauging, how about this:

katrinleinweber:
I though I found a "neat way around this", but then the result table started appearing twice or even thrice for `auto_test()`:
screen shot 2017-07-09 at 12 31 55
¯\\_(ツ)\_/¯ But it works works in both interactive and automatic testing! Now I need to figure out, how to pass the test case to the sourced script.

Issue #66, opened 2017-07-15
Update track metadata as outlined in #65


jonmcalder:
- [x] add a rough estimate of difficulty to each exercise (scale: 1-10)
- [x] add topics to each exercise
- [x] choose at most 20 exercises to be core exercises (set core: true, and delete the unlocked_by key)
- [x] for each exercise that is not core, decide which exercise is the prerequisite (max 1)
- [x] leave 3 or 4 simple exercises as (core: false, unlocked_by: null)

katrinleinweber:
Seeing that #69 deals with the topics, I'll start with the difficulty scale and the core.

jonmcalder:
Cool - I think you should start adding these updates as part of #69 too. I figure it's best to do it all in one branch/pull request since all the changes are within one file.

katrinleinweber:
I initially thought the same, but am now a bit concerned that the "Files changed"-tab will become very crowded. Ah, well, let's try it!

PR #69, opened 2017-07-19
Update track metadata in preparation for nextercism

jonmcalder:
This will address the required updates outlined in #66

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,8 +11,7 @@
     {
       "uuid": "285404a4-d8f8-469d-b183-920b08ba2430",
       "slug": "hello-world",
-      "core": false,
-      "unlocked_by": null,
+      "core": true,
       "difficulty": 1,
       "topics": [
         "text formatting",
@@ -22,8 +21,7 @@
     {
       "uuid": "4be11d9f-db6a-4edd-af35-7d33d2dc9c0f",
       "slug": "leap",
-      "core": false,
-      "unlocked_by": null,
+      "core": true,
       "difficulty": 1,
       "topics": [
         "booleans",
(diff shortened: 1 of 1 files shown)


jonmcalder:
Please do review the topics I added - these are currently based mostly on the other language tracks, with only slight consideration given to the (potential) specifics of the R solutions. They are just there to give us a rough starting point for further discussion.

config.json
```diff
@@ -15,7 +15,8 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "text formatting",
+        "optional values"
```
katrinleinweber: I feel a bit stupid for asking, but are there any _options_ as such in [`hello-world`](https://github.com/exercism/r/blob/master/exercises/hello-world/example.R)? Maybe me being not a native English speaker is the cause for this question. I would use only the 1st topic, plus "strings".

jonmcalder: The `name` argument is optional.

katrinleinweber:
I found this really difficult to decide, because the exercises form just the base of the track. I opted for trying the nextercism progression in a small way.

katrinleinweber:
I understand OP in [/discussions#175](https://github.com/exercism/discussions/issues/175) suggests to be stricter about core exercises and more liberal about unlocking. I'll rework my suggestion.

jonmcalder:
I've updated some of the difficulties and made a slight re-ordering to move bob a little later (since I think both it and beer-song which it is currently unlocking are slightly more tricky than the surrounding exercises. 

See below diagrams visualizing the before and after - nodes are scaled based on difficulty to highlight the progression.
# Before
before
# After
after

katrinleinweber:
Bob was tricky for me mostly because of [an unclear test case](http://exercism.io/submissions/71fa414b1c6f4695b317286cc91488b5). Still, moving it closer to another exercise with similar topics is sensible :-)

jonmcalder:
Cool, thanks! The json is automatically checked by `configlet` too.

PR #73, opened 2017-07-31
Add allergies exercise

jonmcalder:
Add example solution and tests
Add exercise metadata (UUID, difficulty, topics, unlocked_by etc)
Generate README

### DIFF
--- a/exercises/allergies/example.R
+++ b/exercises/allergies/example.R
@@ -1,30 +1,32 @@
+library(magrittr)
+
 allergy <- function(num) {
   
-  allergy_list <- list(
-    cats = 128,
-    pollen = 64,
-    chocolate = 32,
-    tomatoes = 16,
-    strawberries = 8,
-    shellfish = 4,
-    peanuts = 2,
-    eggs = 1
+  allergy_list <- c(
+    "eggs",
+    "peanuts",
+    "shellfish",
+    "strawberries",
+    "tomatoes",
+    "chocolate",
+    "pollen",
+    "cats"
   )
   
   check_allergy <- function(allergy, val) {
-    val %% (allergy * 2) >= allergy
+    intToBits(val)[which(allergy_list == allergy)] %>% as.logical()
   }
   
-  allergies <- lapply(allergy_list, check_allergy, num)
+  allergies <- sapply(allergy_list, check_allergy, num)
   
-  structure(allergies, class = 'allergy')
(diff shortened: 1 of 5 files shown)


exercises/allergies/example.R
```diff
@@ -0,0 +1,34 @@
...
+    chocolate = 32,
+    tomatoes = 16,
+    strawberries = 8,
+    shellfish = 4,
+    peanuts = 2,
+    eggs = 1
+  )
+  
+  check_allergy <- function(allergy, val) {
+    if (val %% (allergy*2) >= allergy) {
```
katrinleinweber: Shouldn't the linter have found the missing spaces around `*`? ;-)

jonmcalder: Well spotted! Turns out I didn't actually include that check in the linter. I should probably add it though. Will do a separate PR for that.

jonmcalder:
Ok no problem - @lorenzwalthert has agreed to have a look at it for us later this month when he's got some time.

exercises/allergies/example.R
```diff
@@ -0,0 +1,30 @@
...
+    chocolate = 32,
+    tomatoes = 16,
+    strawberries = 8,
+    shellfish = 4,
+    peanuts = 2,
+    eggs = 1
+  )
+  
+  check_allergy <- function(allergy, val) {
+    val %% (allergy * 2) >= allergy
```
katrinleinweber: Not a _review comment_, but a good old comprehension question: Is this line a "greedy" check whether any of the given allergy values fits into the total score? In a C tutorial over a year ago, I implemented a greedy algorithm to calculate composing a given amount of change (money) from as few as possible, as large as possible coins. The problem here kind of matches, but I'm having a hard time understanding why this line of math would solve the problem. It does, though O.o

exercises/allergies/example.R
```diff
@@ -0,0 +1,32 @@
...
+    "cats"
+  )
+  
+  check_allergy <- function(allergy, val) {
+    intToBits(val)[which(allergy_list == allergy)] %>% as.logical()
+  }
+  
+  allergies <- sapply(allergy_list, check_allergy, num)
+  
+  structure(allergy_list[allergies], class = 'allergy')
```
lorenzwalthert: use double quotes

jonmcalder: 👍

PR #74, opened 2017-07-31
Update linter

jonmcalder:
Add infix_spaces_linter
Link to tidyverse style guide in README
Update example solutions to comply with the linter update


jonmcalder:
Are there other linters you think we should add? See [lintr README](https://github.com/jimhester/lintr/blob/master/README.md#available-linters). 

My initial thinking was to encourage contributions to follow Hadley's R style guide (which has now become the tidyverse style guide) but to only 'strictly' enforce certain checks via CI, in order to be a little more accommodating towards contributors who may have a slightly different preferences when it comes to code style.

PR #75, opened 2017-08-03
Add crypto-square exercise

jonmcalder:
Implement tests and example solution
Generate README
Update config

### DIFF
--- a/config.json
+++ b/config.json
@@ -335,7 +335,8 @@
         "regular_expressions",
         "transforming"
       ]
-    },
+    }
+  ],
   "foregone": [],
   "language": "R"
 }
--- a/config.json
+++ b/config.json
@@ -314,6 +314,19 @@
         "control_flow_conditionals",
         "recursion"
       ]
+    },
+    {
+      "uuid": "82c43534-86ff-4b5a-9519-1501a15721bf",
+      "slug": "crypto-square",
+      "core": false,
+      "unlocked_by": "rotational-cipher",
+      "difficulty": 4,
+      "topics": [
+        "strings",
+        "text_formatting",
+        "regular_expressions",
+        "transforming"
+      ]
     }
   ]
 }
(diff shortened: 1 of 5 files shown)


jonmcalder:
Following the advice of this discussion in [exercism/problem-specifications](https://github.com/exercism/problem-specifications/pull/250), I opted to include intermediate methods in the tests here. Having said that, I think there might be quite a wide variety of ways to move from the input text to the ciphertext and these intermediate methods might get in the way of an elegant implementation rather than aiding in the learning process. 

My example solution here cleanses the input string and then maps the characters out into a matrix which makes the 'encoding' process easy. This makes the intermediate methods `plaintext_segments` and `encoded` rather pointless though. Obviously an example solution is just an example, but I'm wondering whether maybe we shouldn't drop the these for those two intermediate methods and just keep the first and last methods?

katrinleinweber:
Sorry! I overlooked this the last 2 months, and won't have spare time in the next few weeks either.

jonmcalder:
Ok no worries. I don't think the approach of the example solution matters too much, but I'd like a 2nd opinion on whether these intermediate methods should be included in the starting snippet (and more importantly whether they should be tested), so will just keep this PR open for now until you can revisit it later.

jonmcalder:
Thanks! So you think it's fine to include the tests for the intermediate methods?

katrinleinweber:
> […] maybe we shouldn't drop these for those two intermediate methods and just keep the first and last methods?

I think those prevented me from finding a solution to the last test, because I duplicated `strsplit()` and `paste()` code into `ciphertext` that basically did the opposite of what that code did in the intermediate methods/functions. OTOH I can't spend a few hours again to work out a better example.

Because I think it's better to have a really tricky puzzle in the track, than not, I'm OK with merging this. Please squash, because the two merges & conflict resolutions were this messy, because I tried the web-editor for a change, not a JSON-validating IDE :-/

katrinleinweber:
Hi @jonmcalder! Do you have any doubts about merging & squashing this?

jonmcalder:
Thanks @katrinleinweber! I must have missed/overlooked this somehow - I'm ok with it since you are. I'm sure we (or others) can refine it at a later stage if need be :-)

PR #77, opened 2017-08-05
First stab at intro copy

jonmcalder:
Add blurb to config.json
Update ABOUT.md
Add code snippet

/cc @kytrinyx @robphoenix You guys mentioned that you wouldn't mind proof-reading this content in the related discussion thread?

I'm aware that R as a language often seems very weird to people with an existing programming background, so feedback from those less familiar with R may be especially valuable here.

R isn't really a general purpose programming language, but is also more than just a DSL for data analysis/statistics, since as a computing environment it's highly extensible via it's packaging system and interop with other languages etc.

R definitely has it's quirks though - for example, it supports a mixture of functional and object-oriented programming paradigms: 
- on the functional side of things: it has first class functions, lazy evaluation of arguments, and encourages pure, side-effect free functions (although it doesn't implement tail call recursion and it's easy to create non-pure functions)
- on the OO side of things it now has 3 built in OO paradigms: in two of them objects are immutable and support generic function style OO, while in the third objects are mutable and support the more common message-passing style OO

I've tried to shy away from most of the above in the blurb/about given that I think the idea here is to focus on the selling points of each language, but just thought I'd include this for context?

Any feedback is welcomed :-)

PR #78, opened 2017-08-07
shorten phrases & hyperlink helpful resources

katrinleinweber:
#77 was over a bit too quickly ;-) Here's some rewording that I would like to see in the text.


katrinleinweber:
The "1 sentence/line" was agreed in `meta#84` very shortly before the "[1st stab](https://github.com/exercism/r/pull/77)".

PR #80, opened 2017-10-03
Implement exercise: diamond

jonmcalder:


katrinleinweber:
I finally got around to trying this exercise out. Didn't solve it by myself, but your example is neat! That's good enough for me to include it.

I would like to ask before, whether `nrows` really has to be copied into the function as `rows`. How do you feel about omitting that copy?

jonmcalder:
Oh cool - thanks.

Yes you're right. Looking at it now I don't know why I included that `rows` argument in the function since it is redundant. It should be removed.

jonmcalder:
Thanks @katrinleinweber!

PR #100, opened 2018-06-12
Enable auto_approve for hello-world

katrinleinweber:
I am unsure whether this is necessary, because #98 has a checkbox for this, but later says `We've currently made any hello-world exercises auto-approved in the backend`.

Either way, shall we set other exercises to `"auto_approve": true`?

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,6 +8,7 @@
       "uuid": "285404a4-d8f8-469d-b183-920b08ba2430",
       "core": true,
       "unlocked_by": null,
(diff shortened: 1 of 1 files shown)


jonmcalder:
Thanks for following up on #98! 

I interpret it as saying that `auto_approve: true` should be set on the first exercise if it isn't `hello-world` (and that `hello-world` will be auto-approved by default), but agree that's it's better to be explicit just in case.

I don't think we need/want to auto-approve anything else just yet though.

Issue #103, opened 2018-08-09
Broken link on README


xarxziux:
On the README, under the section [Porting exercises](https://github.com/exercism/r#porting-exercises), the text [list of missing exercises](http://exercism.io/languages/r/todo) is linked to http://exercism.io/languages/r/todo, but this re-directs to https://exercism.io/contribute.  The README needs to be updated to either fix the link, or remove the line if this feature no longer exists.

I've not been able to find any link for missing exercises so I'm assuming that the latter is the case.  Can you please confirm this?

katrinleinweber:
Good catch :-) I believe that the [general porting guide](https://github.com/exercism/docs/blob/798d812dc886a0b6951004cffa072ea649b2b289/you-can-help/implement-an-exercise-from-specification.md#L33) (@kytrinyx?) needs to be updated to align with the new site, so I'd wait for that.

katrinleinweber:
Also, the `porting an exercise`-link can be updated to https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md

kytrinyx:
I have a mountain of work to do in the docs to get everything cleared up. It's going to take a while, I'm afraid.

xarxziux:
Closing this, as the page has been updated.

PR #104, opened 2018-08-09
Update broken links in README in accordance with issue #103

xarxziux:
As discussed in issue #103, I've removed the line referring to the `list of missing exercises` and updated the `porting an exercise` link.


katrinleinweber:
Thank you! Let's wait for 48h for an update from above to #103 ;-)

PR #108, opened 2018-08-30
difference-of-squares: renamed square_sums to square_sum

Scientifica96:
Renamed `square_sums` to `square_sum`.


Scientifica96:
It's a pleasure :D

Issue #109, opened 2018-09-05
Unit tests should check for vector parameters


xarxziux:
Having completed all the core exercises and most of the side exercises on the R track, it's become clear to me that while Exercisms testing model works well enough for general-purpose languages, it breaks down for languages that do things a little differently like R.  The general design of the tests is that they assume that the student is going to produce a function that accepts a single value.  To test several input values, the student's function is called several times with different single values.

This works fine for a general-purpose language like Python or JavaScript, but it breaks down for R which has no concept of single values.  Instead, a declaration like `x <- 1` does not create an integer or number value, but a numeric vector of size 1.  This is a very important concept in R and one that needs to be emphasised to students of the language.  I didn't have a firm grasp of this concept myself until recently, despite using R in a professional capacity for some time.  Unfortunately the unit tests as they stand don't require the student to "think in R" and allow them to assume that the input values are single-element vectors.  I've come to realise while working through the track that I've been guilty of this myself.

The solution to this is quite straightforward: in each exercise, in addition to the current tests, there should be one final test that takes all the values from the previous tests, combines them into a vector (or list, if appropriate) and sends that vector or list to the students function.  This will check if they've made the mistake of assuming that the input is a single value, or if they've respected the R way of doing things.  This may well break existing solutions (including most of mine), but I think it's a change worth making.

I'm happy to submit the PR's for this myself, but I'd like to see if you agree with my observations and think this is worth doing first.

jonmcalder:
Thanks for the feedback!

This is an interesting point that you raise and so I get where you are coming from. I'm not entirely sure I agree with the conclusion that you jumping to here though (i.e. to extend all the tests to explicitly check for this).

> This is a very important concept in R and one that needs to be emphasised to students of the language.

I agree that this is important an important concept, but I also want to be sensitive to the spectrum of users of R. Often newer users of R are not necessarily aiming to be R developers (although that may change over time). Maybe for the audience of Exercism users learning/practicing R they will typically tend to be more towards the developer end of the spectrum (especially if they know other programming languages already, but this is not always the case).

My gut feel is that maybe this concept of things in R generally being vectorized by default should be made explicit later on in the R track and thus only introduced into the later tests? Alternatively maybe even just making these optional extra tests to pass (maybe with motivations in one or two key exercises as to why this concept is important/valuable for R developers).

@katrinleinweber, @jkanche & @amoradell what are your thoughts?

katrinleinweber:
[...]

PR #111, opened 2018-10-06
Add vectorised unit test for Collatz Conjecture

xarxziux:
As a demonstration of the point I raised in issue #109, here is an update for the non-core Collatz Conjecture exercise.  The current version of this does not test for vector parameters and the example would fail this test.

The following changes have been made:
- An extra test has been added to the unit tests that send a four-element vector to the function.
- In the example, the original function `collatz_step_counter` has been renamed to `collatz_scalar`.
- Also in the example, the line `collatz_step_counter <- Vectorize(collatz_scalar)` has been added.

This is a very minor change to the example, but makes the function a lot more powerful by taking full advantage of R's vectorisation capabilities.  R developers really shouldn't be thinking in single values the way most mainstream programming languages do.

### DIFF
--- a/exercises/collatz-conjecture/example.R
+++ b/exercises/collatz-conjecture/example.R
@@ -1,13 +1,15 @@
-collatz_step_counter <- function(num) {
-  
+collatz_scalar <- function(num) {
+
   if (num <= 0) {
     stop("Only positive numbers are allowed")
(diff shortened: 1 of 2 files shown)


katrinleinweber:
I have this on the radar, and will try to review it this weekend. Sorry for the delay!

PR #115, opened 2019-03-03
Deduplicate mentor note

katrinleinweber:
This accompanies https://github.com/exercism/website-copy/pull/828. Let's discuss whether the (newer) example code from the mentor note makes a better example.

cc @exercism/r

### DIFF
--- a/exercises/sum-of-multiples/example.R
+++ b/exercises/sum-of-multiples/example.R
@@ -1,16 +1,4 @@
 sum_of_multiples <- function(factors, limit) {
-  
-  get_mults <- function(factor) {
-    
-    max_mult <- ceiling(limit / factor) - 1
-    if (max_mult > 0) {
-      new_multiples <- factor * 1:max_mult
-    }
-    
-  }
-  
-  multiples <- unique(unlist(lapply(factors, FUN = get_mults)))
-  
-  return(sum(as.numeric(multiples)))
-  
+  func <- function(x) seq(0, limit - 1, x)
+  sum(unique(unlist(lapply(factors, func))))
 }
--- a/exercises/sum-of-multiples/example.R
+++ b/exercises/sum-of-multiples/example.R
@@ -1,4 +1,16 @@
 sum_of_multiples <- function(factors, limit) {
-  func <- function(x) seq(0, limit - 1, x)
-  sum(unique(unlist(lapply(factors, func))))
+  
+  get_mults <- function(factor) {
+    
+    max_mult <- ceiling(limit / factor) - 1
+    if (max_mult > 0) {
+      new_multiples <- factor * 1:max_mult
+    }
+    
+  }
+  
+  multiples <- unique(unlist(lapply(factors, FUN = get_mults)))
+  
+  return(sum(as.numeric(multiples)))
+  
 }


jonmcalder:
The idea of linking to the example code from the mentor notes seems like a good one.

And the proposed solution here is certainly much more concise and readable. It is worth noting however, that it seems to be less performant by roughly a factor of 20 for lengthy factor vectors (see below). I'd be happy to accept this as is, but maybe this performance/readability tradeoff is worth noting in the mentor notes?

I haven't really given much thought to this problem or explored what a 'good' solution for this is - so there may be a different approach that's better than both of these.

```
sum_of_multiples <- function(factors, limit) {
  get_mults <- function(factor) {   
    max_mult <- ceiling(limit / factor) - 1
    if (max_mult > 0) {
      new_multiples <- factor * 1:max_mult
    }    
  }  
  multiples <- unique(unlist(lapply(factors, FUN = get_mults)))  
  return(sum(as.numeric(multiples)))
}

sum_of_multiples2 <- function(factors, limit) {
  func <- function(x) seq(0, limit - 1, x)
  sum(unique(unlist(lapply(factors, func))))
}

library(microbenchmark)

microbenchmark(
  sum_of_multiples(1:1000,10),
  sum_of_multiples2(1:1000,10)
)

Unit: microseconds
                          expr       min         lq      mean    median        uq      max neval
  sum_of_multiples(1:1000, 10)   946.723   986.3395  1104.015  1007.277  1031.295  7983.91   100
 sum_of_multiples2(1:1000, 10) 20015.815 20829.7245 22101.764 21530.939 22990.639 44784.12   100
```

katrinleinweber:
In that case, we don't have to change the example at all ;-)

I mostly didn't want to drop the newer one silently.

SleeplessByte:
I don't think this needs to be closed persé; the discussion here is very valuable. See https://github.com/exercism/website-copy/pull/828#issuecomment-469447926 for why the other PR was closed.

katrinleinweber:
I'm failing to extract generally useful [comments from community solutions](https://exercism.io/tracks/r/exercises/sum-of-multiples/solutions?utf8=%E2%9C%93&order=num_comments). Please feel free to take over adding some comments.

katrinleinweber:
Maybe the [`completed` `Sum of Multiples` exercises that each of us mentored](https://exercism.io/mentor/dashboard/your_solutions) are a better source for comments & notes?

jonmcalder:
Thanks for persevering with this @katrinleinweber!

To answer your earlier question:
> In that case, let's leave the example code as is, but maybe add a few comments to explain the efficiency (it's ceiling(limit / factor), right?).

The main reason the example solution is more efficient is because `seq()` is an S3 generic - so there is a (slight) bit of overhead in using it due to the method dispatch. This overhead is non-negligible when calling the function repeatedly (within `lapply`). You can check this by replacing `seq()` with `seq.int()`:

```
microbenchmark(
  sum_of_multiples_seq(1:1000,100),
  sum_of_multiples_seq_int(1:1000,100)
)

Unit: milliseconds
                                  expr       min        lq      mean    median        uq       max neval
     sum_of_multiples_seq(1:1000, 100) 20.707143 21.403223 23.175072 21.677880 23.730201 82.729626   100
 sum_of_multiples_seq_int(1:1000, 100)  1.224661  1.287063  1.366065  1.320318  1.349056  3.797145   100
```

In terms of general comments for mentoring, I think the current mentor notes already cover the key things, namely: 
- that `for` loops should be avoided in favour of the `apply` or `map` function families due to R's interpreted nature, and 
- that this exercise is great one to introduce packages like `microbenchmark` and discuss the merits of experimentation to assess the impacts of different functional choices on performance.

PR #120, opened 2019-05-08
rotational-cipher: Prevent solutions that simply duplicate the alphabet

katrinleinweber:
As discussed in https://github.com/exercism/problem-specifications/pull/1438 and other issues/PRs linked there, I would like to suggest that the R track adds test for rotation keys <0 and >26. What do you think?

Things I am unsure about:

- [ ] [+1 the difficulty](https://github.com/exercism/r/blob/rotate-more-thoroughly/config.json#L164)?
- [ ] bump exercise version?

### DIFF
--- a/exercises/rotational-cipher/README.md
+++ b/exercises/rotational-cipher/README.md
@@ -4,7 +4,7 @@ Create an implementation of the rotational cipher, also sometimes called the Cae
 
 The Caesar cipher is a simple shift cipher that relies on
 transposing all the letters in the alphabet using an integer key
-between `0` and `26`. Using a key of `0` or `26` will always yield
+(both positive or negative). Using the same key will always yield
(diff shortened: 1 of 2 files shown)


depial:
Closing this due to this exercise being in `problem-specifications` and having been updated since this PR was opened.

From a practical standpoint, while I liked the idea of preventing solutions that duplicate the alphabet, this fix only discourages that because a solution with alphabet duplication need only redefine `key <- key %% 26` to pass the proposed tests.

@colinleach, feel free to reopen if you disagree, but I think the benefits of sticking to `problem-specifications` outweigh the benefits of adding these extra tests.

colinleach:
No disagreement from me. I always apposed deviations from `problem-specifications`, and that is even more important now we have templates.

PR #126, opened 2019-08-12
Add config for pre-commit hooks and update README

jonmcalder:
Seeing the number of linting messages being returned by lintr bot in some of the recent PR's made me realize that pre-commit hooks could prove pretty useful here (assuming contributors can be encouraged to install them).

This PR adds [pre-commit](https://pre-commit.com) config from https://github.com/lorenzwalthert/pre-commit-hooks (`styler`/`parsable-r`) and updates the README with some setup instructions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+- repo: https://github.com/lorenzwalthert/pre-commit-hooks
+  rev: v0.0.0.9015
+  hooks: 
+  -   id: style-files
+  -   id: parsable-R
+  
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+- repo: https://github.com/lorenzwalthert/pre-commit-hooks
+  rev: v0.0.0.9015
```
katrinleinweber: Do we need to be strictly use only that revision? Would `master` be a problem? We could omit the mention of `autoupdate` then.

jonmcalder: I think pre-commit configuration works this way by design - see https://pre-commit.com/#using-the-latest-version-for-a-repository

Maybe @lorenzwalthert can confirm / clarify this though?

lorenzwalthert:
I believe you need an exact version yes. Mentioning the auto update mechanism sounds like a good idea.

katrinleinweber:
OK, cool :-) Thanks for putting this together!

PR #127, opened 2019-08-24
Adjust core track to help ease bottlenecks

kytrinyx:
**TL;DR:**
This PR reorders the core exercises; see the end for the new ordering.
Please read the full thing before replying!

---

I wrote a script that takes some data points into account to move certain core exercises to be optional side exercises, and which reorders the remaining exercises.

This is not meant to be the "be all end all" of perfection :-)
This is an attempt to make some simple changes in the short term. In the longer term, over the next 12 months, the product team will be doing a bunch of work to dig into what makes great Exercism exercises, and how to structure tracks to provide a better experience for both learners and mentors.

**Once this goes live** we will continue to monitor the continuation rate, the median wait time, and a few other stats as well, to help decide whether or not we need to take any other immediate steps. That said, if you see anything weird or worrying once the change is out, please ping us on Slack (or here) so we can discuss a fix posthaste!

**The most helpful thing you can do in reviewing this** is to look at the results and tell me whether anything is obviously terrible or suspiciously weird. I know very little about the R track (or the language, for that matter), so I could easily have missed something.

I'd also appreciate it if you posted this to the mentors as a heads up, and also because mentors tend to have a great gut sense about the core exercises.

### Removal from core:

The following exercises were moved out of the core track:

- sum-of-multiples
- largest-series-product
- pascals-triangle

This is a list of exercises that typically are unappealing to students and not very interesting to mentor. They tend to be math problems or implementations of CS algorithms and the like.

**Worth considering:** Are any of these exercises particularly valuable in the R track? If so, we should keep it. Let me know and I can regenerate the PR.

### Reordering:

The data that I based the reordering on was taken from the production database, and is based on the last 6 months.

I reordered by taking the original order, and then giving "penalty points" for two different things

- low continuation rates (percentage)
- long wait times in the mentor queue (median wait time in minutes)

"Continuation" is when someone completes an exercise, but then does not submit the following exercise. We have other interesting numbers that we may use to adjust things in the future, but this one seemed like our best bet for now for the bottlenecks. I somewhat arbitrarily chose 75% as the cutoff for whether or not to give penalty points, i.e. continuation rate < 75% gets penalized. The only exercise I did not penalize based on this rate was the exercise directly following `hello-world`, since `hello-world` is basically just a taste test, and it is natural to assume that many people will decide that Exercism is not really for them.

<details>
 <summary>Continuation rates</summary>

```
- hello-world (100%)
- two-fer (69%)
- leap (89%)
- hamming (82%)
- raindrops (88%)
- bob (73%)
- anagram (80%)
- sum-of-multiples (100%)
- word-count (77%)
- phone-number (100%)
- rotational-cipher (78%)
- largest-series-product (83%)
- pascals-triangle (100%)
```

</details>

Long wait times in the mentor queue is an indication that mentors don't enjoy mentoring the exercise, or put it off. This could be for a number of reasons, which we have not explicitly identified, but we've observed that sometimes the exercise is simply too difficult in the current position, which means that mentors have to go back and forth a whole bunch with students to get something right, and the discussions can be quite frustrating. Also sometimes the exercise is just not very interesting to mentor. Or if the exercise is too early in the track sometimes people will submit lots of really complicated solutions rather than a small handful of mostly reasonable solutions, also making it harder to mentor.

So this pushes exercises with long wait times further back in the core track. We may decide that we need to remove some exercises from core altogether, if wait times continue to be bad.

<details>
 <summary>Wait times</summary>

```
- hello-world (0 min)
- two-fer (5666 min)
- leap (478 min)
- hamming (5868 min)
- raindrops (2937 min)
- bob (27236 min)
- anagram (8240 min)
- sum-of-multiples (31500 min)
- word-count (7600 min)
- phone-number (3229 min)
- rotational-cipher (11229 min)
- largest-series-product (4260 min)
- pascals-triangle (2591 min)
```
</details>

#### Before
```
- hello-world
- two-fer
- leap
- hamming
- raindrops
- bob
- anagram
- sum-of-multiples
- word-count
- phone-number
- rotational-cipher
- largest-series-product
- pascals-triangle
```

#### After

_Note: the numbers indicate the position in the core track, as defined by the index of the array. Minus means that it moved earlier, plus means that it moved later._
```
- hello-world
- two-fer
- leap
- raindrops (-1)
- hamming (+1)
- phone-number (-3)
- anagram
- word-count
- rotational-cipher (-1)
- bob (+4)
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -38,17 +38,6 @@
       ]
     },
     {
-      "slug": "hamming",
-      "uuid": "dcc7171a-6b7d-4749-a300-a7270a095486",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-        "filtering",
-        "strings"
-      ]
-    },
-    {
       "slug": "raindrops",
       "uuid": "955604ef-266f-42c9-abde-ce0fdc2d33ad",
       "core": true,
@@ -60,14 +49,25 @@
       ]
     },
     {
-      "slug": "bob",
-      "uuid": "da01ce65-8c7d-4646-ace4-c067d92f656b",
+      "slug": "hamming",
+      "uuid": "dcc7171a-6b7d-4749-a300-a7270a095486",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "filtering",
+        "strings"
+      ]
+    },
+    {
+      "slug": "phone-number",
+      "uuid": "e5166378-e3fc-4393-b4f8-4c36da444c56",
       "core": true,
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
-        "control_flow_conditionals",
-        "strings"
+        "parsing",
+        "transforming"
       ]
     },
     {
@@ -93,23 +93,23 @@
       ]
     },
     {
-      "slug": "phone-number",
-      "uuid": "e5166378-e3fc-4393-b4f8-4c36da444c56",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 2,
-      "topics": [
-        "parsing",
-        "transforming"
-      ]
-    },
-    {
       "slug": "rotational-cipher",
       "uuid": "8bde91c5-2fe7-4f8d-91ec-ed05284b9a49",
       "core": true,
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
+        "strings"
+      ]
+    },
+    {
+      "slug": "bob",
+      "uuid": "da01ce65-8c7d-4646-ace4-c067d92f656b",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": [
+        "control_flow_conditionals",
         "strings"
       ]
     },
--- a/config.json
+++ b/config.json
@@ -128,7 +128,7 @@
       "slug": "difference-of-squares",
       "uuid": "6a562684-5938-4ad9-b042-7b44a8cc4561",
       "core": false,
-      "unlocked_by": null,
+      "unlocked_by": "leap",
       "difficulty": 1,
       "topics": [
         "integers",
--- a/config.json
+++ b/config.json
@@ -103,6 +103,17 @@
       ]
     },
     {
+      "slug": "tournament",
+      "uuid": "acf63670-7e56-47c1-9f33-7ec076d5270f",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "parsing",
+        "text_formatting"
+      ]
+    },
+    {
       "slug": "bob",
       "uuid": "da01ce65-8c7d-4646-ace4-c067d92f656b",
       "core": false,
@@ -278,17 +289,6 @@
       "topics": [
         "control_flow_conditionals",
         "math"
-      ]
-    },
-    {
-      "slug": "tournament",
-      "uuid": "acf63670-7e56-47c1-9f33-7ec076d5270f",
-      "core": true,
-      "unlocked_by": "rotational-cipher",
-      "difficulty": 4,
-      "topics": [
-        "parsing",
-        "text_formatting"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -105,8 +105,8 @@
     {
       "slug": "bob",
       "uuid": "da01ce65-8c7d-4646-ace4-c067d92f656b",
-      "core": true,
-      "unlocked_by": null,
+      "core": false,
+      "unlocked_by": "anagram",
       "difficulty": 2,
       "topics": [
         "control_flow_conditionals",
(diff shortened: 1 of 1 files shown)


katrinleinweber:
I trust you & that analysis, and thus tend to approve. However:

- [ ] want to check the `topic` order

before merging.

kytrinyx:
I ran `configlet fmt .` which I _think_ checks topic order?

Second thing: if you think we should keep the mathy exercises in this track, I trust your judgement on that. The "let's not have math exercises" is more of a general thing, and if R is mathy folks, mainly, then it might not be a problem here.

config.json
```diff
@@ -215,7 +215,7 @@
       "slug": "prime-factors",
       "uuid": "ee849e23-d709-4e4a-ab89-07a87e13e4fb",
       "core": false,
-      "unlocked_by": "sum-of-multiples",
+      "unlocked_by": null,
```
katrinleinweber: This might be a bit of an issue. We anchored 5 optional exercises to `sum-of-multiples`. Those become free-floating with this change, plus some others.

katrinleinweber:
### After

Screen Shot 2019-08-25 at 21 06 37

This many free-floating ones has me worried. I'll add some suggestions to re-anchor as many exercises as possible.

### Before

Screen Shot 2019-08-25 at 21 15 17

katrinleinweber:
OK, this optional-izes most of the math exercises early, after `leap`.

Screen Shot 2019-08-25 at 21 56 36

jonmcalder:
Thanks for this @kytrinyx, and thanks @katrinleinweber for reviewing and suggesting some re-configuration of the track to eliminate the many floating exercises that would result from removing the mathy / algorithmic exercises from the core track.

I love the data-driven approach here, though I'm a little skeptical whether we've really got enough critical mass (both in terms of mentors & students) for these waiting times and continuation rates to be reliable indicators.

That being said, the thinking seems reasonable. The only thing that feels a bit wrong to me is moving `bob` right to the end of the core track. Obviously the R track is still rather under-developed, but even so, the core exercises should at least progress in difficulty and `bob` is far from the hardest of the core exercises here so far.

I think it should surface a bit earlier on but be removed from the core track given that it seems to be a slight bottleneck. I'd suggest that it instead be unlocked by `anagram`. `beer-song` should move with it or maybe be unlocked by the following core exercise (`word-count`) if we want to spread things out a little more.

In the process of re-thinking the relative difficulty of some of these exercises and looking at e.g. the python and go tracks, I realized that `tournament` in particular is a tougher exercise which should be moved much later, and it's a core exercise in both of those tracks too so I think it could become our final core exercise for now.

That's my two cents.

kytrinyx:
> eliminate the many floating exercises that would result from removing the mathy / algorithmic exercises from the core track.

Oooh, yes, I forgot to mention that in the PR description. I couldn't make any good automated guesses about the free-floating ones, so you did exactly what I was hoping, making reasoned decisions based on your knowledge of the track 💙 Thank you for that!

> I'm a little skeptical whether we've really got enough critical mass (both in terms of mentors & students) for these waiting times and continuation rates to be reliable indicators.

Yeah, that's a good point. There are tracks that I'm explicitly not doing data-driven stuff with, because the level of activity is just too low. I *think* that we have enough activity here to give an indicator, but that is worth double-checking. @iHiD -- would you take a look at the raw numbers for the R track since Jan 1st to see what the order fo magnitude is?

> The only thing that feels a bit wrong to me is moving bob right to the end of the core track.

Bob is not difficult, but it has a lot more complexity than I originally realized, which has led it to be moved back in every track I've looked at the data for, so far.

> I think it should surface a bit earlier on but be removed from the core track given that it seems to be a slight bottleneck.

I like this suggestion! I was about to suggest putting it right after `anagram`, but this is even better.

>  I realized that tournament in particular is a tougher exercise which should be moved much later

Yeah, that would make sense! Do you have any other side exercises that you think would make sense to move into the core track at this point?

jonmcalder:
Thanks for the follow-up and feedback @kytrinyx!

> Yeah, that would make sense! Do you have any other side exercises that you think would make sense to move into the core track at this point?

Based on some of the other tracks, I was thinking that `isogram`, `scrabble-score`, and `luhn` are possible candidates for the core track here, but I feel a little out of touch at the moment - did the exercises so long ago and haven't been mentoring that much lately.

katrinleinweber:
Yeah, same here. @jonmcalder: Can you please try to implement the `config.json` changes discussed above? I'm having a confusing problem with the visualizer which blocks me from checking the changes when I implement them.

iHiD:
We have a cohort of 146 students, so it's a small dataset. But the patterns (e.g. bob) are consistent across all tracks. (Bob is the worst exercise in basically every track, due to its complexity and large solution-space, and therefore its pain to mentor).

I'd personally trust the continuation rates, but less so the wait-times.

jonmcalder:
Thanks @iHiD for the follow-up. That makes sense.

@katrinleinweber I've updated the config per earlier discussion (bob to side-track and unlocked by anagram, and tournament moved to core towards the end of the track).

Any further suggestions or should we leave it like this for now?

image

PR #136, opened 2020-11-09
Move to Github Actions

### LINKED ISSUE #135: Moving from Travis to GitHub Actions
Hello 🙂 

Over the last few months we've been transferring all our CI from Travis to [GitHub Actions (GHA)](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions). We've found that GHA are easier to work with, more reliable, and much much faster. 

Based on our success with GHA and increasing intermittent failures on Travis, we have now decided to try and remove Travis from Exercism's org altogether and [shift everything to GHA](https://github.com/exercism/exercism/issues/5396). This issue acts as a call to action if your track is still using Travis. 

For most CI checks this should be a transposing from Travis' syntax to [GHA syntax](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions), and hopefully quite straightforward (see [this PR for an example](https://github.com/exercism/csharp/pull/1338/files)).  However, if you do encounter any issues doing this, please ask on Slack where lots of us now have experience with GHA, or post a comment here and I'll tag relevant people. This would also make a good Hacktoberfest issue for anyone interested in making their first contribution 🙂 

If you've already switched this track to GHA, please feel free to close this issue and ignore it.

Thanks!

jonmcalder:
Closes #135


jonmcalder:
@exercism/github-actions

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Install project dependencies
+        run: |
+          install.packages(c("testthat", "jsonlite", "lintr", "magrittr"))
+        shell: Rscript {0}
+
+      - name: Run exercism/r ci pre-check (checks config, lint code) for all exercises
+        run: |
+          source("bin/run_lints.R")
+        shell: Rscript {0}
```
ErikSchierboom, line 46: Was there a specific reason not to run the linting in the `ci` step below? If it was integrated there, it would be a bit faster due to not having to do the checkout, setup of R and install package steps being doubled here.

ErikSchierboom (approved):
LGTM

.github/workflows/ci.yml
jonmcalder: Only did it this way cos I was working from the template suggested here: https://github.com/exercism/problem-specifications/blob/master/workflows/ci.yml

But I guess the steps were separated in the template cos linting usually only needs to be checked on a single stable version (and possibly has fewer dependencies) whereas one generally wants to run the tests on multiple versions (and more of the dependencies are required here) in case of any incompatibilities?

Is that tradeoff worth it or do you think I should rather just combine the lint and test checks in the CI step?

ErikSchierboom: Ah right. It probably makes sense to adhere to the default workflow then. I missed that, sorry.

jonmcalder: Cool no worries. Thanks so much for the review!

ErikSchierboom:
@jonmcalder Do you have merge permissions or do you want me to merge?

jonmcalder:
All good thanks!

ErikSchierboom:
Thanks for doing the PR!

PR #185, opened 2021-09-02
Improve instructions and tests for Anagram exercise

stragu:
This PR fixes a few things:

* The instructions a clearer, as they now make it obvious that:
   * The output needs to be a vector (not a list)
   * The match needs to be case-insensitive
   * Same words shouldn't be matched
* The tests are now more flexible: they check for a length of 0 for empty vectors, instead of an equality to `c()`. This fixes the issue of users submitting a valid solution that returns an empty vector of mode `character` and having tests failing because of that.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.md
+++ b/exercises/practice/anagram/.docs/instructions.md
@@ -1,8 +1,10 @@
 # Instructions
 
 An anagram is a rearrangement of letters to form a new word.
-Given a word and a list of candidates, select the sublist of anagrams of the given word.
+Given a word and a vector of candidates, return the subset of anagrams of the given word.
 
-Given `"listen"` and a list of candidates like `"enlists" "google"
-"inlets" "banana"` the program should return a list containing
-`"inlets"`.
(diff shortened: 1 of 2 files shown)


jonmcalder:
Thanks for the PR and sorry for the delay in following up!

I'm happy with the changes to the tests you propose and will accept those.

Regarding the instructions: these are generally not meant to be track specific (by design) and are automatically synced from the [problem specifications](https://github.com/exercism/problem-specifications/) repo.

e.g. the use of the term `list` is incidental and shouldn't imply the list datatype in R (which is rather unfortunate I know)

In general I think there is a bit of an expectation that students should be able to learn to deduce what's required from the tests themselves (as that's a valuable part of the learning process), but that being said if you're suggesting these changes to the instructions because of things that tripped you up and you feel that there should be further clarification in the instructions then the place to add these is in `.docs/instructions.append.md` as outlined [here](https://exercism.org/docs/building/tracks/practice-exercises).

Could I ask you to please update your PR to expand on / clarify the existing instructions rather than modifying them directly?

And thanks again - I appreciate you contributing to the R track!

PR #192, opened 2022-01-10
Ensure division and derived operations aren't used

b4D8:
- [x] Add a test for that purpose
- [x] Modify instructions accordingly


jonmcalder (changes requested):
Hi there, and thanks for this PR. 

In order to consider this, the change / addition to the instructions should be included in `instructions.append.md` as outlined in the [practice exercises documentation](https://exercism.org/docs/building/tracks/practice-exercises).

The reason for this is because for existing problems shared across multiple tracks (like this one), the content of `instructions.md` is automatically sync'd via configlet based on the content outlined in the problem-specifications repo.

b4D8:
Hi, thanks for your feedback.
Sorry I didn't take time to actually read this documentation..
I'm closing this and will create another PR.

PR #193, opened 2022-01-12
[sieve] Test that provided solution doesn't use division operation

b4D8:
Hi, 

I think it makes sense to ensure division was not used to solve this exercise.

Note that I updated the test so that it only uses base R.

### DIFF
--- a/exercises/practice/sieve/test_sieve.R
+++ b/exercises/practice/sieve/test_sieve.R
@@ -37,11 +37,11 @@ test_that("find primes up to 1000", {
                  967, 971, 977, 983, 991, 997))
 })
 
-test_that("division not used", {
-  division_ops <- c("/", "%", "sqrt", "\\^", "exp")
-  division_ops <- paste(division_ops, collapse = "|")
-  use_division <- any(grepl(division_ops, deparse(sieve)))
-  expect_false(use_division)
-}
+# test_that("division not used", {
+#   division_ops <- c("/", "%", "sqrt", "\\^", "exp")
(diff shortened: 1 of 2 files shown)


exercises/practice/sieve/test_sieve.R
```diff
@@ -37,4 +37,11 @@ test_that("find primes up to 1000", {
                  967, 971, 977, 983, 991, 997))
 })
 
+test_that("division not used", {
+  division_ops <- c("/", "%", "sqrt", "\\^", "exp")
+  division_ops <- paste(division_ops, collapse = "|")
+  use_division <- any(grepl(division_ops, withVisible(sieve)$value))
```
jonmcalder, line 43: This doesn't work because `withVisible(sieve)$value` returns a function (closure) and not a string.

I don't see why `withVisible` is necessary though, surely one could simply use `deparse(sieve)`?

jonmcalder (changes requested):
Current test results in an error - please address.

I'd also like to request that this additional test be commented out so that it's not required by default but can be enabled by those who wish to try and adhere strictly to the algorithm.

Thanks!

exercises/practice/sieve/test_sieve.R
b4D8: You're right! Sorry, I must have mixed things up while trying to remove the dependency to `utils` by replacing `capture.output`. `deparse` will perfectly do the job. Thanks!

jonmcalder (approved):
Thanks!

PR #202, opened 2022-01-25
:construction: Lasagna concept exercise

jonmcalder:
First concept exercise


ErikSchierboom (commented):
Excellent! Do you want to continue working on this in this PR or want to work with follow-up PRs?

jonmcalder:
@ErikSchierboom Happy to carry on working in this PR a little further first thanks.

Hoping to get a review from @b4d8 and then update the docs. Will ping you for a review when ready to merge (but obviously will keep exercise status WIP in the config for a while until we have planned out and implemented a few more concept exercises later in the year and can hopefully move some to BETA).

b4D8:
Hi @jonmcalder, 

Great! I double checked the text and results. Everything looks good to me.
Here are some observations:

* I had a look at how the different tracks implement this and noticed a divergence in how the value of `expected_minutes_in_oven` is declared. A function is also used in Rust (const is explicitly out of scope) while Javascript, Python and Go rely on constants. Julia doesn't expect it to be declared at all. I was thinking we could also store the value in a variable rather that wrap it in a function. This would demonstrate the fact that variables are accessible from within a function (related to scoping rules and environments without naming or talking about those concepts directly). I don't have a strong opinion on this though.

* It doesn't seem really important but I also noticed that the actual value for `expected_minutes_in_oven` is set to 40 for Rust, Javascript, Go, Elixir and Python (also 60 for julia).. 🤷‍♂️

* Should we work on adding the following concepts: `function`, `variables`, `arithmetic`? We could then use a little summary as an introduction to the exercise (like [julia](https://github.com/exercism/julia/blob/main/exercises/concept/lasagna/.docs/introduction.md)?

* I also like the way [julia adds hints](https://github.com/exercism/julia/blob/main/exercises/concept/lasagna/.docs/hints.md). Even if R doesn't have an official online documentation for people working on the website, we could refer to [rdocumentation.org](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/Arithmetic) and `?"+"` for cli users.

* As for the next concept exercise, interestingly, also in julia, this exercise is followed by [documented-lasagna](https://github.com/exercism/julia/tree/main/exercises/concept/documented-lasagna) which asks the student to document the lasagna's functions. While it might seem inappropriate to introduce roxygen so early, it also give room to explaining how to request help `?` and use the documentation which I think could be of use for the student's journey. I prefer this approach to Python's which seems to force students to document all functions (in this exercise at least) which I think could be confusing. I we were to follow this road, using a function for `expected_minutes_in_oven` as it is would probably be less confusing.

ErikSchierboom:
> I had a look at how the different tracks implement this and noticed a divergence in how the value of expected_minutes_in_oven is declared. A function is also used in Rust (const is explicitly out of scope) while Javascript, Python and Go rely on constants. Julia doesn't expect it to be declared at all. I was thinking we could also store the value in a variable rather that wrap it in a function. This would demonstrate the fact that variables are accessible from within a function (related to scoping rules and environments without naming or talking about those concepts directly). I don't have a strong opinion on this though.

Different tracks make different decisions. For example, in C# I chose _not_ to use a constant, as I should then also explain what the difference between const and readonly is. It would also require more syntax to be learnt (the `const` keyword), which was not desirable for an introduction exercise that already introduces quite a bit. If it is nothing more than a variable in R, you could definitely consider it.

> It doesn't seem really important but I also noticed that the actual value for expected_minutes_in_oven is set to 40 for Rust, Javascript, Go, Elixir and Python (also 60 for julia).. 🤷‍♂️

Hmmm, that's odd :D 

> Should we work on adding the following concepts: function, variables, arithmetic? We could then use a little summary as an introduction to the exercise (like julia?

You could, but you don't have to. Most tracks have just one `basics` concept. This has two advantages:

1. It becomes easier to add as a prerequisite for other Concept Exercises (you just have to list `basics`, instead of having to list `functions`, `variables`, `arithmetic`).
2. You can have separate concepts for functions, arithmetic (and possibly variables) where you go much more in-depth , which is probably something you want anyway.

My recommendation would be to just have one `basics` concept.

> I also like the way julia adds hints. Even if R doesn't have an official online documentation for people working on the website, we could refer to rdocumentation.org and ?"+" for cli users.

Sounds like a good plan.

> As for the next concept exercise, interestingly, also in julia, this exercise is followed by documented-lasagna which asks the student to document the lasagna's functions. While it might seem inappropriate to introduce roxygen so early, it also give room to explaining how to request help ? and use the documentation which I think could be of use for the student's journey. I prefer this approach to Python's which seems to force students to document all functions (in this exercise at least) which I think could be confusing. I we were to follow this road, using a function for expected_minutes_in_oven as it is would probably be less confusing.

Is documentation an important part of R? If not, you don't _have_ to have a comments exercise (which is usually rather dull anyway). I think most tracks don't have a separate comments exercise, and just explain what comments are in the `basics` exercise.

jonmcalder:
Thanks @b4D8 and @ErikSchierboom - very helpful!

> Different tracks make different decisions. For example, in C# I chose _not_ to use a constant, as I should then also explain what the difference between const and readonly is. It would also require more syntax to be learnt (the `const` keyword), which was not desirable for an introduction exercise that already introduces quite a bit. If it is nothing more than a variable in R, you could definitely consider it.

I'll consider switching to a variable then, as it is really straightforward in R. My only concern would be that it kinda implicitly introduces the idea of scoping (R uses lexical scoping) which we'd want to ignore until a little later on in the learning track. I assume that's ok?

> My recommendation would be to just have one `basics` concept.

Thanks - will stick with `basics` for now and look to possibly go deeper into `functions` etc later on.
 
> Is documentation an important part of R? If not, you don't _have_ to have a comments exercise (which is usually rather dull anyway). I think most tracks don't have a separate comments exercise, and just explain what comments are in the `basics` exercise.

In R formal documentation usually involves adding skeleton comments adhering to a particular format which are used to generate docs in the officially supported format via an additional R library (`roxygen`). This is not required and usually only done when creating an R package so I think we should just stick with comments in the basic exercise and maybe introduce formal documenting of functions later on.

ErikSchierboom:
> My only concern would be that it kinda implicitly introduces the idea of scoping (R uses lexical scoping) which we'd want to ignore until a little later on in the learning track. I assume that's ok?

Yes. The first couple of Concept Exercises (and especially `basics`) can be a bit hand-wavey and gloss over these details. You would then likely have a separate exercise where you introduce scoping.

> This is not required and usually only done when creating an R package so I think we should just stick with comments in the basic exercise and maybe introduce formal documenting of functions later on.

👍

colinleach (approved):
As discussed in PR #202

colinleach (approved):

PR #213, opened 2022-08-12
expect_is --> expect_equal

KaiAragaki:
It's unclear why `expect_is` is used here - it will yield cryptic errors when correct solutions are passed. `expect_is` (as the documentation says) is more for type-comparison and is now deprecated anyway.

### DIFF
--- a/exercises/practice/triangle/test_triangle.R
+++ b/exercises/practice/triangle/test_triangle.R
@@ -4,7 +4,7 @@ library(testthat)
 context("triangle")
 
 test_that("true if all sides are equal", {
-  expect_is(triangle(2, 2, 2), "equilateral")
+  expect_equal(triangle(2, 2, 2), "equilateral")
 })
 
 test_that("false if any side is unequal", {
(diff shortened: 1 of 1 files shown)


jonmcalder:
Thanks for picking this up and for the PR @KaiAragaki!

The test for type-comparison is actually intentional, since the idea with this exercise is to implement a simple S3 class and have "types" of triangles rather than just a function which returns a string descriptor. If you don't feel the exercise instructions are clear enough in that regard could you suggest an improvement [here](https://github.com/exercism/r/blob/main/exercises/practice/triangle/.docs/instructions.append.md)?

These tests were written before {testthat} v3 and the deprecation of `expect_is()` - they should definitely be updated to use `expect_s3_class()` so as to be more clear & explicit. Assuming you agree could you please make that change in this PR?

KaiAragaki:
@jonmcalder Whoops! I didn't expect that reading the hint would have been necessary for solving this problem. Reading prompts is apparently not my strong suit. I do find it a bit odd to introduce classes in this problem (particularly for an easy problem), since R tends to be fairly functional (except when it comes to package development - at least in my travels).

I'll close this PR, perhaps open a new one regarding `expect_is`/`expect_s3_class` at a later time. 

Thank you for your reply!

PR #229, opened 2023-03-16
✨ Add ETL practice exercise

jonmcalder:


jonmcalder:
My example solution here seems a little arcane, but I couldn't come up with anything better 🤷‍♂️

ErikSchierboom (approved):
Great!

PR #232, opened 2023-03-24
fizz-buzz: add tests and improve instructions

ErikSchierboom:
I felt like the tests could maybe be a bit more gradual in ramping up the difficulty.
I also slightly rephrased the instructions.
Let me know what you think.


ErikSchierboom:
@jonmcalder Fair point! I've updated

jonmcalder:
Thanks!

PR #233, opened 2023-04-02
test(anagram): clean up redundant tests

ofhope:
Going through this exercise I notice a fair bit of redundant testing. They loose value making it harder to establish the expected return.

* Removed multiple tests for _detects multiple anagrams_
* Removed multiple tests for _anagrams must use all letters exactly once_
* Removed multiple test for case insensitive checking
* Clarified test descriptions

### DIFF
--- a/exercises/practice/anagram/test_anagram.R
+++ b/exercises/practice/anagram/test_anagram.R
@@ -10,34 +10,6 @@ test_that("no matches", {
                c())
 })
 
-test_that("detects simple anagram", {
-  subject <- "ant"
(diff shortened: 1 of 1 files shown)


exercises/practice/anagram/test_anagram.R
```diff
@@ -10,25 +10,26 @@ test_that("no matches", {
                c())
 })
 
-test_that("detects simple anagram", {
```
ofhope, line 13: There is no distinction between simple and complex anagrams. Removing this adjective.

PR #242, opened 2023-04-16
darts: add practice exercise

colinleach:
A simple one, to make sure I understand the process.

### DIFF
--- a/config.json
+++ b/config.json
@@ -72,6 +72,19 @@
         ]
       },
       {
+        "slug": "darts",
+        "name": "Darts",
+        "uuid": "ddbd717d-a6b5-40fc-9dc8-3526e7409091",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "status": "active",
+        "topics": [
+          "control_flow_conditionals",
+          "math"
+        ]
+      },
+      {
         "slug": "raindrops",
         "name": "Raindrops",
         "uuid": "955604ef-266f-42c9-abde-ce0fdc2d33ad",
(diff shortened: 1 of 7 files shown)


colinleach:
One for @ErikSchierboom

config.json
```diff
@@ -485,6 +485,19 @@
...
         ]
+      },
+      {
+        "slug": "darts",
+        "name": "Darts",
+        "uuid": "ddbd717d-a6b5-40fc-9dc8-3526e7409091",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "status": "wip",
```
ErikSchierboom, line 496: I think it's fine to merge this with `active`

```suggestion
        "status": "active",
```

ErikSchierboom (commented):
Some minor suggestions. I'll leave the actual R review to @jonmcalder.

exercises/practice/darts/.meta/config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [
+  ],
```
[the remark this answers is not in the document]
colinleach: done

config.json
```diff
@@ -485,6 +485,19 @@
           "loops",
           "math"
         ]
+      },
```
jonmcalder, line 488: Can we please move this earlier (I'd suggest just after leap) to facilitate recommended order on the website?

jonmcalder (changes requested):
I think this should probably be one of the first few recommended practice exercises, but other than that this PR is great as is - thanks again!

FWIW I bashed out a quick solution in the process of reviewing of this PR and it ended up being the same (literally verbatim) as your examplar solution 😄

config.json
colinleach: Changed as requested

jonmcalder (approved):

PR #243, opened 2023-04-16
armstrong-numbers practice exercise

colinleach:
I think we probably need to omit the last 2 tests (commented out for now), which use integers of 108 and 127 bits respectively: far beyond R's int32 default. Nothing I've seen on StackOverflow is encouraging, and the suggested workarounds are probably unsuitable for Exercism testing servers.

We may need @ErikSchierboom to reopen this PR.

### DIFF
--- a/config.json
+++ b/config.json
@@ -485,6 +485,18 @@
           "loops",
           "math"
         ]
+      },
+      {
+        "slug": "armstrong-numbers",
+        "name": "Armstrong Numbers",
+        "uuid": "c17ef56b-7418-428b-ac9d-37dbd28de32e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "status": "wip",
+        "topics": [
+          "math"
+        ]
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -492,8 +492,8 @@
         "uuid": "c17ef56b-7418-428b-ac9d-37dbd28de32e",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 1,
-        "status": "wip",
+        "difficulty": 2,
+        "status": "active",
         "topics": [
           "math"
         ]
--- a/exercises/practice/armstrong-numbers/.meta/example.R
+++ b/exercises/practice/armstrong-numbers/.meta/example.R
@@ -1,0 +1,9 @@
+to_digits <- function(n) {
+  chars = n |> as.character() |> strsplit("")
+  as.numeric(chars[[1]])
+}
+
+is_armstrong_number <- function(n) {
+  digits = to_digits(n)
+  sum(digits ^ length(digits)) == n
+}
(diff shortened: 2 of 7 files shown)


jonmcalder:
Thanks for these PRs @colinleach! I will try and review sometime tomorrow or otherwise as soon as I find a gap 👍

config.json
```diff
@@ -485,6 +485,18 @@
...
         ]
+      },
+      {
+        "slug": "armstrong-numbers",
+        "name": "Armstrong Numbers",
+        "uuid": "c17ef56b-7418-428b-ac9d-37dbd28de32e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "status": "wip",
```
ErikSchierboom, line 496:
```suggestion
        "status": "active",
```

exercises/practice/armstrong-numbers/.meta/example.R
```diff
@@ -0,0 +1,9 @@
+to_digits <- function(n) {
+  chars <- n |> as.character() |> strsplit("")
+  as.numeric(chars[[1]])
+}
+
+is_armstrong_number <- function(n) {
+  digits <- to_digits(n)
+  sum(digits ^ length(digits)) == n
+}
```
ErikSchierboom, line 9: Love this implemementation!

exercises/practice/armstrong-numbers/test_armstrong-numbers.R
```diff
@@ -0,0 +1,50 @@
...
+  expect_equal(is_armstrong_number(9926314), FALSE)
+})
+
+# test_that("Armstrong number containing seven zeroes", {
+#   expect_equal(is_armstrong_number(186709961001538790100634132976990), TRUE)
+# })
+# 
+# test_that("The largest and last Armstrong number", {
+#   expect_equal(is_armstrong_number(115132219018763992565095597973971522401), TRUE)
+# })
```
ErikSchierboom, line 48: Why are these tests commented?

colinleach: These fail in R (and many other languages) because of integer overflow. I think they need to be deleted but I wanted to give an R expert a chance to disagree.

Deleting also gets rid of the linter error.

ErikSchierboom: Okay makes sense. If they _are_ excluded, remember to add `include = false` to the corresponding entries in the `tests.toml` file: https://github.com/exercism/r/pull/243/files#diff-28bb739e7bc3f68b1a554f3abf5553f2a3423fe5f9d0951b68b7130b1b11cb40

jonmcalder (approved):
Thanks @colinleach! I agree with excluding the tests for large integers and am happy with this implementation.

jonmcalder:
The examplar solution fails on R v3.6.3 because the |> operator was only introduced in R v4.1.0, but happy to ignore this for now and maybe we should stop testing on older versions of R at some point?

colinleach:
Oops! I've been using |> in various exercises (after doing a bunch of F# in Functional February!), so we'll see this problem again.

PR #244, opened 2023-04-17
Queen attack: new practice exercise

colinleach:
Some questions:
- Is `list` a good way to implement the queen's position? There are various other options.
- Is it a good idea to check for specific error messages? I see that older exercises only check for a generic error, but I don't know why.

I'm creating this PR now, but you may want to delay review until I've had feedback on earlier PRs. I assume I still have plenty to learn about this process, to make life easier for the reviewers.

**Edit Apr 20th:** Think more about this, I dislike my first test  - too implementation specific. The Python track just tests for no error when creating a valid queen, ignoring the return value. Maybe we should change to `expect_error(... , NA)` for a valid result?


jonmcalder (changes requested):
Nothing wrong with using a list to implement the Queen's position, but I think I'd prefer to test only the outcomes i.e. simply expect_error() for create() with invalid positions, and true/false for can_attack(). If learners aren't restricted to a specific implementation then learning and mentoring opportunities are more open ended which I see as a good thing as long as the instructions and expectations (tests) are clear which I think they are.

colinleach:
I already tested this change locally, so it's an easy PR to send. Give me a few minutes - I'm currently in the middle of a minor domestic situation.

jonmcalder:
Oh, and re checking for specific error messages vs a generic error, I'm open to be persuaded othewise, but have generally felt that a lighter touch for error checking is better for most exercises since the different possible error states either aren't meaningfully different, or could presumably easily have been differentiated if they have each been detected / checked reliably.

Or in other words, requiring a certain error message or custom error class expectation seems like it would add unnecessary friction (especially for earlier exercises when other fundamental learning objectives should be the focus). But again, I'm open to suggestions on how error testing could / should be done better if it's not consistent or rigorous enough. Maybe for practice exercises beyond a certain point, or with higher difficulty, error expectations should be more strict?

exercises/practice/queen-attack/queen-attack.R
```diff
@@ -0,0 +1,7 @@
+create <- function(row, col) {
+
+  }
```
jonmcalder, line 3:
```suggestion
}
```

config.json
```diff
@@ -485,6 +485,19 @@
...
         ]
+      },
+      {
+        "slug": "queen-attack",
+        "name": "Queen Attack",
+        "uuid": "50a29cbf-a782-4d71-9127-94d110f3c893",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "status": "wip",
```
jonmcalder, line 496:
```suggestion
        "status": "active",
```

colinleach:
I think I got all the changes in, including removal of specific error messages: those were taken from `canonical-data.json` but I don't have strong views either way about including them.

exercises/practice/queen-attack/test_queen-attack.R
```diff
@@ -0,0 +1,90 @@
...
+library(testthat)
+
+context("queen-attack")
+
+# Test creation of Queens with valid and invalid positions
+
+test_that("queen with a valid position", {
+  row <- 2
+  col <- 2
+  expect_error(create(row, col), NA)
```
jonmcalder, line 11:
```suggestion
  expect_error(create(row, col))
```

jonmcalder (approved):
Thank you!

PR #248, opened 2023-04-22
Clock: new practice exercise

colinleach:
Another one to leave auto-closed for now

Several tracks agree difficulty 3


colinleach:
Deleting because this exercise was created by another author.

PR #251, opened 2023-06-12
test(anagram): clean up redundant tests

ofhope:
I tried to open this a few months ago during a hiatus. Reopening again 🤞🏽 https://github.com/exercism/r/pull/233

Going through this exercise I notice a fair bit of redundant testing. They loose value when duplicated and make it harder to establish the expected result.

Removed multiple tests for "detects multiple anagrams"
Removed multiple tests for "anagrams must use all letters exactly once"
Removed multiple test for "case insensitive checking"
Clarified test descriptions

### DIFF
--- a/exercises/practice/anagram/test_anagram.R
+++ b/exercises/practice/anagram/test_anagram.R
@@ -10,34 +10,6 @@ test_that("no matches", {
                c())
 })
 
-test_that("detects simple anagram", {
-  subject <- "ant"
-  candidates <- c("tan", "stand", "at")
-  expect_equal(anagram(subject, candidates),
-               c("tan"))
-})
-
-test_that("does not detect false positives", {
-  subject <- "galea"
-  candidates <- c("eagle")
-  expect_equal(anagram(subject, candidates),
-               c())
-})
-
-test_that("detects multiple anagrams", {
-  subject <- "master"
-  candidates <- c("stream", "pigeon", "maters")
-  expect_equal(anagram(subject, candidates),
-               c("stream", "maters"))
-})
-
-test_that("does not detect anagram subsets", {
-  subject <- "good"
-  candidates <- c("dog", "goody")
-  expect_equal(anagram(subject, candidates),
-               c())
-})
-
 test_that("detects anagram", {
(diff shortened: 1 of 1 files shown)


jonmcalder (changes requested):
Thanks.

If you'd like to update the tests for this exercise, then please align the changes to match the tests outlined in the [problem specifications repo](https://github.com/exercism/problem-specifications/blob/main/exercises/anagram/canonical-data.json)

Track tests do fall out of sync with what's specified (as has happened here), but we generally try not to deviate from those tests unless there is a track-specific or language-specific reason to do so.

ofhope:
I think I'll close this down. It looks like the specification has duplication meaning these tests would need to reflect that.

Test cases are immutable with only description, comments or scenario additions being allowed.

> Test cases are immutable, which means that once a test case has been added, it never changes. 

So every language track needs to implement redundant tests to align with this problem specification. It makes sense to make them immutable so the many dependant language tracks don't fall out of alignment and aren't routinely having to sync. But the OCD in me has a little eye twitch while reading this specification 😄 

Duplicate test example (the candidates input are slightly different but from a test perspective I would call them identical).

```
{
  "uuid": "a0705568-628c-4b55-9798-82e4acde51ca",
  "description": "words other than themselves can be anagrams",
  "property": "findAnagrams",
  "input": {
    "subject": "LISTEN",
    "candidates": [
      "Listen",
      "Silent",
      "LISTEN"
    ]
  },
  "expected": [
    "Silent"
  ]
},
{
  "uuid": "33d3f67e-fbb9-49d3-a90e-0beb00861da7",
  "reimplements": "a0705568-628c-4b55-9798-82e4acde51ca",
  "description": "words other than themselves can be anagrams",
  "property": "findAnagrams",
  "input": {
    "subject": "LISTEN",
    "candidates": [
      "LISTEN",
      "Silent"
    ]
  },
  "expected": [
    "Silent"
  ]
}
```

ofhope:
Oh... wait I just noticed the field  `reimplements` and see the section "Changing Tests".  I'll take a closer look at this specification and see if I can align it.

PR #256, opened 2023-07-05
vehicle-purchase concept exercise

colinleach:
For `conditionals`. Concept will follow in the next PR.


exercises/concept/vehicle-purchase/.docs/introduction.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+In this case, parentheses `()` around the boolean and braces `{}` around the statement are both required. 
+A simple `if` statement can be shortened:
+
+```R
+if (x == 0) return("success")
+```
+
+An alternative if-else form may be useful:
```
jonmcalder, line 37: The fact that ifelse is vectorised is an important caveat here. I think it's better not to mention it here at all since it's not necessary or useful for this exercise, and it is addressed in about.md

exercises/concept/vehicle-purchase/.docs/hints.md
```diff
@@ -0,0 +1 @@
+# Hints
```
jonmcalder, line 1: Maybe worth mentioning `min` here to help with `choose_vehicle()`?

(I assume you're avoiding `sort()` since then one has to address vector subsetting).

exercises/concept/vehicle-purchase/.meta/design.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Goal
+
+The goal of this exercise is to teach the student about conditionals in R.
+
+## Learning objectives
+
+- Know that the relational operators include `==`, `!=`, `>=`, `<=`.
+- Know the syntax of a block if-else.
+- Know about the `ifelse()` function.
```
jonmcalder, line 12: If use of `ifelse()` with vectors is out of scope for this exercise (which I agree with), then I don't agree with "knowing about the `ifelse()` function" being one of the learning objectives.

exercises/concept/vehicle-purchase/.meta/exemplar.R
```diff
@@ -0,0 +1,17 @@
+needs_license <- function(vehicle) {
+  vehicle %in% c("car", "truck")
```
jonmcalder, line 2: Seeing that both here and in the implementation for `choose_vehicle()` you've made use of `c()` makes me wonder whether maybe it would be better if the vectors concept were introduced before conditionals?

Of course both `needs_license()` and `choose_vehicle()` can be implemented using scalar values plus if/else, so that could be done here but it wouldn't be idiomatic R code.

jonmcalder (changes requested):
I'm happy with the implementation here, but I'm not quite convinced about putting this conditionals exercise before the vectors one because I think idiomatic R solutions for this exercise are likely to involve some use of vectors. Maybe that's not a dealbreaker though, I just think it may be worth reconsidering.

I see that the exemplar solutions for the vectors concept exercise do make use of `if()`, so if conditionals came after vectors then we'd have to navigate that issue somehow. You've probably given this more thought than I have so let me know what you think.

exercises/concept/vehicle-purchase/.docs/introduction.md
colinleach: removed

exercises/concept/vehicle-purchase/.meta/design.md
colinleach: changed

exercises/concept/vehicle-purchase/.docs/hints.md
colinleach: Depends on the main discussion, about which order the exercises are presented.

colinleach:
Every track has a nightmare with cyclic dependencies in the learning syllabus as they try to sequence exercises, and this is also true here. My thinking, rightly or wrongly, was that the first 3 concepts follow a sequence that students are familiar with from other tracks, then in concept 4 (`vectors`) we start to move to idiomatic R. I cheated a bit in `exemplar.R` and used `c()`, but this isn't essential.

I just re-read the Exercism documentation on [syllabus design](https://exercism.org/docs/building/tracks/syllabus). They explicitly encourage hand-waving:

_"Sometimes you'll feel like there is a deadlock. Concept A requires understanding concept B, and B requires understanding A._

_In this case simplify. Hand-wave over some complexity in one so that you can get students familiar with the other. It's perfectly fine to say that something will be introduced in more depth later, and that for now the student just needs to understand this one bit._

_Concepts are understood more deeply in stages and over time."_

That could, of course, be used as an argument for conditionals before _or_ after vectors!

I'm normally uncomfortable with evading making a decision (former research manager speaking), but might this be a good point to flag up the issue and solicit opinions from other maintainers? In preparing teaching materials, how they are perceived by others is pretty much the point.

colinleach:
Thinking about this a bit more, maybe it's the exercise that is wrong for the concept, rather than concept sequencing?
Would `conditionals` before `vectors` work better if `cars-assemble` introduced it (#264)?

jonmcalder:
Cool - points well made and I understand that some deadlocks will be unavoidable. 

I agree that it probably makes sense to introduce the first few concepts in a similar order to what many other tracks do, so maybe deferring vectors slightly is the right call.

`cars-assemble` feels slightly cleaner and better to me for introducing the conditionals concept because it doesn't involve vectors. The exercise is obviously more mathematical (which might be perceived as more difficult), but since R is well suited to numeric problems I think in general it may make sense to lean into these sorts of examples to illustrate concepts.

Would you be happy to swap out `cars-assemble` for `vehicle-purchases`?

colinleach:
Absolutely, no problem.

exercises/concept/vehicle-purchase/.meta/exemplar.R
colinleach: I've pushed a new version of `exemplar.R` that removes all use of vectors. Is it horribly unidiomatic? `needs_license` gets a bit longer and less elegant, but `choose_vehicle` is slightly shorter if given multiple arguments without wrapping them in `c()`.

I also extended `introduction.md` to mention `max()` and `min()` with multiple arguments. If you're happy with this, we'd need to propagate it to `concepts/conditionals`.

colinleach: Alternatively, to keep `%in% c()` we could add `vectors` as a prerequisite for this exercise. It involves some jumping around the concept tree, but I think that's pretty normal for exercises that are not the first in their concept.

exercises/concept/vehicle-purchase/.meta/exemplar.R
```diff
@@ -0,0 +1,20 @@
+needs_license <- function(vehicle) {
+  if (vehicle == "car" || vehicle == "truck") {
+    return(TRUE)
+  }
+  FALSE
```
jonmcalder, line 5:
```suggestion
  vehicle == "car" || vehicle == "truck"
```
The if / else is unnecessary here.

colinleach: Obviously, yes. I'm embarrassed to have missed that one - clearly having a bad day.

exercises/concept/vehicle-purchase/.meta/exemplar.R
jonmcalder: I don't think we need to make vectors a prerequisite. `needs_license` seems ok because there are only 2 options - it would be worse if there were multiple. So happy to keep as is for now.

jonmcalder (approved):

PR #257, opened 2023-07-05
Conditionals

colinleach:
Concept corresponding to `vehicle-purchase`.


colinleach:
`about.md` goes a bit beyond `introduction.md` and mentions use of vectors in conditionals. I didn't go as far as including `logical()`, `as.logical()` or `is.logical`. Perhaps too early in the syllabus for those? I could make an argument for or against.

concepts/conditionals/about.md
```diff
@@ -0,0 +1,59 @@
...
+A simple `if` statement can be shortened:
+
+```R
+if (x == 0) return("success")
+```
+
+## Using vectors in conditionals
+
+Looking ahead to the `vector-filtering` concept: the conditional clause in parentheses `()` must evaluate to a single `TRUE`/`FALSE`. 
+Including vectors in the comparison usually gives a vector of booleans, so these will need to be wrapped in an aggregating function such as `all()` or `any()`.
```
jonmcalder, line 42: I think it would be helpful to follow this up with an example to illustrate.

e.g. 
```
numbers <- c(4,5,7,9,10,11)
if (any(numbers %% 3 == 0)) print("1 or more numbers are divisible by 3")
```

(maybe you can think of a better example)

concepts/conditionals/about.md
```diff
@@ -0,0 +1,59 @@
...
+
+## The `ifelse` function
+
+An alternative if-else form may be useful, providing there is only only a true/false outcome.
```
jonmcalder, line 46:
```suggestion
An alternative if-else form may be useful, provided there is only a true/false outcome.
```

jonmcalder (changes requested):
Please make the requested changes.

concepts/conditionals/about.md
colinleach: No disagreement. It's a bad time now, but I'll make these changes tomorrow.

jonmcalder:
> `about.md` goes a bit beyond `introduction.md` and mentions use of vectors in conditionals. I didn't go as far as including `logical()`, `as.logical()` or `is.logical`. Perhaps too early in the syllabus for those? I could make an argument for or against.

I don't see the need to address these at this early stage, so leave as is for now, and we can always update later.

concepts/conditionals/about.md
colinleach: I've added a slightly extended version of this to `about.md`

colinleach:
While adding jonmcalder to the contributor list, I noticed the blurb still mentioned `ifelse()` - now removed

jonmcalder (approved):

PR #262, opened 2023-07-08
elyses-analytic-enchantments concept exercise

colinleach:
First exercise for the `vector-filtering` concept. I'm a bit conflicted about the concept name: filtering is probably clearer for a general audience, subsetting is maybe more technically correct.


colinleach:
Apparently, only one exercise per concept can be a concept exercise, any others have to be practice exercises. We'll need to pick either this or `bird-watcher` #267.

PR #267, opened 2023-07-08
bird-watcher concept exercise

colinleach:
Second exercise for `vector-filtering`, after `elyses-analytic-enchantments`.


colinleach:
Apparently, only one exercise per concept can be a concept exercise, any others have to be practice exercises. We'll need to pick either this or `elyses-analytic-enchantments` #262.

Issue #252, opened 2023-07-05
Progress towards Learning Syllabus


colinleach:
Concept taxonomy was discussed in [issue #198](https://github.com/exercism/r/issues/198). This issue is about implementation, borrowing extensively from the earlier issue.

A [tracking spreadsheet](https://docs.google.com/spreadsheets/d/1xW5WDL5Xp2yQ9ZFJY-oVj1c19kNkZLwCiSEjH1H2SGI/edit#gid=0) for Concepts and Concept Exercises is available. A flowchart of how it might fit together is shown below.

## Some that are already in progress or planned

This is a very fluid plan and will doubtless continue to change frequently!

- Boxes in solid red will almost certainly need a new concept exercise created. 
- Others marked `??` with red borders can probably fork something suitable from other tracks (TODO).
- Those with green borders have had PRs submitted for both the exercise and the associated concept.
- Those in solid green have had both merged.

```mermaid
graph TD
    
start(("Start")) --basics--> basics["lasagna"]
style basics fill:#0f0

basics --booleans--> booleans["annalyns-infiltration"]
style booleans fill:#0f0

booleans --conditionals--> conditionals["cars-assemble<br/>(vehicle-purchase)"]
style conditionals fill:#0f0

conditionals --vectors--> vectors["elyses-enchantments"]
style vectors fill:#0f0

switch -.loops.-> loops["mixed-juices"]
style loops stroke:#0f0

conditionals --switch--> switch["blackjack"]
style switch stroke:#0f0

vectors --set-operations--> set-operations["ozans-playlist"]
style set-operations stroke:#0f0

name-attribute --lists--> lists["need-for-speed"]

vectors --vector-filtering--> vector-filtering["elyses-analytic-enchantments<br/>bird-watcher"]
style vector-filtering stroke:#0f0

vector-filtering --name-attribute--> name-attribute["??"]
style name-attribute fill:#f00,color:#fff

vector-filtering --vector-functions--> vector-functions["??"]
style vector-functions stroke:#f00

vector-filtering --nothingness--> nothingness["name-badges"]
style vector-filtering stroke:#0f0

vectors --strings--> strings["log-levels"]
style strings stroke:#0f0

strings --nothingness--> nothingness["name-badges"]
style nothingness stroke:#0f0

strings --randomness--> randomness["captains-log"]
style randomness stroke:#0f0

strings --regular-expressions--> regular-expressions["regular-chatbot"]

lists --lapply-sapply--> lapply-sapply["??"]
[...]

PR #270, opened 2023-07-10
log-levels concept exercise

colinleach:
This one feels messy and unsatisfactory, but for now it's the best I've got. Because of all the Docker-related difficulties in adding `tidyverse` to the test runner earlier this year, I've stuck to core R functions while dreaming of `stringr`.

A number of other concepts have this as a prerequisite, but all most of them really need is `paste()` and/or `sprintf()`.

Any thoughts on the best way forward?


colinleach:
See issue #252.

PR #274, opened 2023-07-11
randomness concept

colinleach:
Only really covers `sample()` and `runif()`. The `about.md` mentions the large number of other options but without detail.


colinleach:
See issue #252.

PR #275, opened 2023-07-13
blackjack concept exercise

colinleach:
Exercise for the `switch` concept, though it also practices `conditionals` covered previously.


colinleach:
See issue #252.

PR #281, opened 2023-07-18
names-attribute concept

colinleach:
I don't yet have an exercise to match this concept. I'll try to add one in the next few days.


colinleach:
See issue #252.

PR #284, opened 2023-07-21
errors concept

colinleach:
I've provisionally called this `errors`, but alternative names include `exceptions` or `conditions`.

The corresponding exercise is `factory-sensors`. I've done most of that, but need to add a final part to illustrate `warnings` rather than `errors`.


colinleach:
See issue #252.

PR #286, opened 2023-07-27
move `vehicle-purchase` from concept to practice exercise.

colinleach:
It turns out having 2 concept exercises for the same concept breaks the syllabus. We have `cars-assemble` as the concept exercise for `conditionals`, so `vehicle-purchase` needs to be a practice exercise.

I've left it as `wip`, to stop `configlet lint` complaining that there is no student-facing concept for the `booleans` prerequisite.


jonmcalder (approved):
I had noticed the concept duplication in the syllabus but forgot to follow up so thanks for addressing this!

iHiD:
Hey all. 

So we don't support this. Once an exercise is in a config.json, it's in for good. You can't delete it or change its type). It has to be deprecated (with the appropriate flag/status). 

In this case I've managed to change data in the database manually to put things in an ok state, but this involved deleting Colin's solution. This wouldn't be a doable fix if we had more solutions, but it's fine for this situation.

So no harm done, but probably worth just double checking with Erik when doing anything "unusual" like this :)

---

@ErikSchierboom I wonder if configlet or some CI should catch this? Both removing exercises and changing the type should be banned.

jonmcalder:
Ah ok that makes sense. Apologies, and thank you for the manual DB fixing and for clarifying @iHiD!

colinleach:
Oops, sorry! I knew you don't like re-purposing practice exercises as concept, but I (wrongly) thought it would be OK to move a wip concept exercise that had been wrongly classified in the first place, and that almost no students had even seen. I stand corrected...

Thank you to you and Erik for untangling this mess!

PR #303, opened 2024-01-06
Add reverse-string

colinleach:
For week 2 of #48in24, as discussed.

No approaches yet. Once this one is merged, I'll do that separately and PR it for discussion. The first one is often the most trouble.

### DIFF
--- a/config.json
+++ b/config.json
@@ -593,6 +593,14 @@
         "topics": [
           "math"
         ]
+      },
+      {
+        "slug": "reverse-string",
+        "name": "Reverse String",
+        "uuid": "684e330d-7ee6-4151-9fc3-c1ca376136c3",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
       }
     ]
   },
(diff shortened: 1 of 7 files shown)


jonmcalder (approved):
Thanks - this looks great!

For the example solution, my preference would be:

```
reverse <- function(text) {
  strsplit(text, "") |>
    lapply(rev) |>
    sapply(paste, collapse = "")
}
```

because it'll handle vector text input, and because (these days) I think use of the native pipe is often preferable to assigning / naming intermediate objects (at least for shorter chains of operation - though this is probably a question of style / readability rather than what's idiomatic in R).
 
Not a big deal, so happy for you to leave this as is (i.e. change it only if you wish to do so after seeing this).

But another reason I thought it worth mentioning is because it could be worth exploring vectorised vs non-vectorised solutions in one or more of the performance articles. I'm considering doing so for leap. 

(in the above case and in general, depending on implementation details, non-vectorised solutions will be faster for single inputs and slower for multiple inputs)
```

colinleach:
Your version is, of course, much better. I changed it.

PR #305, opened 2024-01-14
[Reverse String]: add approaches

colinleach:
Draft for discussion. We have about 9 days before this exercises launches on #48in24, not bad for the (rather chaotic!) start of a new Exercism annual challenge.

I had intended to include a performance article, but that's behind schedule. I have done enough to include summary results in the `approaches` documents, but code, graphs, etc will probably take another couple of days (and be very non-finalized, especially the graphs). Given the plethora of files involved, splitting the PRs may be a good strategy in any case.

### DIFF
--- a/exercises/practice/reverse-string/.approaches/introduction.md
+++ b/exercises/practice/reverse-string/.approaches/introduction.md
@@ -120,7 +120,7 @@ See the [`vector-operations`][vector-operations] approach for further details.
 
 ## Performance
 
-Full details will be given in the [Performance article][performance-article], with just a summary here.
+Full details will be given in a Performance article (to follow soon), with just a summary here.
 
 Testing each of the approaches about 10,000 times with a single 5-character string gave the timing data below.
 
--- a/exercises/practice/reverse-string/.approaches/introduction.md
+++ b/exercises/practice/reverse-string/.approaches/introduction.md
@@ -1,0 +1,176 @@
+# Introduction
+
+There are surprising many ways to solve this exercise.
+We will start with simple approaches, considering some variations in syntax, then go on to look at vector operations and run-time performance.
+
+## General considerations.
+
+Base R is relatively weak on string handling.
+Strengths include:
+- Vector handling, so we could convert the string to a vector, reverse that, then convert back to a string.
+- Libraries, that can be used to provide functionality beyond base R.
+
+## Use split-reverse-paste
+
+```r
+reverse <- function(text) {
+  splits <- strsplit(text, "")[[1]]
+  reversed <- rev(splits)
+  paste(reversed, collapse = "")
+}
+```
+
+This is a relatively simple approach, that has worked since early versions of R.
+
+The first example uses intermediate variables.
+We could instead use nested function calls: more concise, perhaps less readable.
+
+```r
+reverse <- function(text) {
+  paste(rev(strsplit(text, "")[[1]]), collapse = "")
+}
+```
+
+See the [`split-reverse`][split-reverse] approach for further details.
+
+From R 4.1.0 (introduced May 2021), the base language also has native pipes, allowing a syntax more typical of functional languages:
+
+```r
+reverse <- function(text) {
+  strsplit(text, "")[[1]] |>
+    rev() |>
+    paste(collapse = "")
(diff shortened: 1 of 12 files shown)


exercises/practice/reverse-string/.approaches/introduction.md
```diff
@@ -0,0 +1,176 @@
...
+
+There are surprising many ways to solve this exercise.
+We will start with simple approaches, considering some variations in syntax, then go on to look at vector operations and run-time performance.
+
+## General considerations.
+
+Base R is relatively weak on string handling.
+Strengths include:
+- Vector handling, so we could convert the string to a vector, reverse that, then convert back to a string.
+- Libraries, that can be used to provide functionality beyond base R.
```
jonmcalder, line 11:
```suggestion
Base R provides a reasonable set of string operations, but these have grown organically over time and can be a little inconsistent / hard to learn. Often string manipulation in base R will require the use of regular expressions and by contrast to other high-level languages like Python or Ruby some string manipulation can be a little harder to do in (base) R.

However, packages such as [{stringr}](https://stringr.tidyverse.org/) and the underlying [{stringi}](https://stringi.gagolewski.com/index.html) are easy to use, expressive and efficient for manipulating strings in a myriad of ways. In practice many developers will turn to libraries such as these for manipulating strings, especially when needing to do more complex string processing.

But to begin with, bearing in mind that vectors are right at the heart of R (more on that later), a natural approach to reverse a string is to convert the string to a vector of characters, reverse that, and then convert back to a string.
```

(feel free not to accept this suggestion if it is a little verbose, or to borrow parts of it if you like only some of it - maybe I'm just biased and defensive but I feel a statement like "string handling in base R is relatively weak" needs a little more qualification.

exercises/practice/reverse-string/.approaches/native-pipes/snippet.txt
```diff
@@ -0,0 +1,5 @@
+reverse <- function(text) {
+  strsplit(text, "") |>
+    lapply(rev) |>
+    sapply(paste, collapse = "")
```
jonmcalder, line 4:
```suggestion
    rev() |>
    paste(collapse = "")
```
(I think this should be the non-vectorised version since the vectorised version is a separate approach?)

jonmcalder (changes requested):
This is great - thanks so much @colinleach.

I've made a few suggestions for what they are worth, but otherwise am largely happy with this as is.

Keen to get this merged sooner rather than later given that the exercise is being featured next week and we can always amend and/or make improvements later. I think the placeholder reference to the performance article will need to be removed unless you're willing to include (the beginnings of) that at this stage.

exercises/practice/reverse-string/.approaches/introduction.md
```diff
@@ -0,0 +1,176 @@
+# Introduction
+
+There are surprising many ways to solve this exercise.
```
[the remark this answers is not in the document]
colinleach: Of course, I meant _surprisingly_, but yours is good too.

colinleach:
> I think the placeholder reference to the performance article will need to be removed unless you're willing to include (the beginnings of) that at this stage.

I've started the performance article, so (health willing!) I hope to get a first draft out soon-ish. However, I can't promise we'll have something finished very soon, so it may be safer to remove the references to it.

colinleach:
I'm a bit wary about promising much on the performance article, because there are arguments on multiple tracks about the best way to graph the data. `ggplot2` certainly gives us plenty of options!

I didn't know until a few days ago that graphics can't be included in this repo, they need to go to another one that syncs with S3 to support the website. Also, they have to be done for both light and dark backgrounds. Also (preferably) there are a lot of accessibility guidelines that we are encouraged to follow.

I think the first one may hit delays. Fingers crossed that we get better at this as the year progresses...

jonmcalder:
Yes there is definitely a fair amount involved in including graphs for both light and dark mode and trying to do a good job following accessibility guidelines. Completely happy with this as is for now and thanks again for your work on this! 🤗 

Also sorry that you've been sick - hope you bounce back soon!

jonmcalder (approved):

PR #310, opened 2024-01-22
[Roman Numerals] add practice exercise

colinleach:
The performance stuff for `reverse-string` is going badly, so I did this in the interim. It's for week [02-06]. Only the exercise, no Deep Dive stuff.

It works, but the example implementation may not be the most idiomatic R. A `while` loop inside a `for` loop is standard across most languages (except the ones relying heavily on pattern matching), but I'll be very happy to be shown that I'm missing something R-specific.

Feel free to replace with something better if you wish: we can use all the implementations for approaches later. It certainly isn't vectorized, but I couldn't see how to make that work within the function.

### DIFF
--- a/config.json
+++ b/config.json
@@ -601,6 +601,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1
+      },
+      {
+        "slug": "roman-numerals",
+        "name": "Roman Numerals",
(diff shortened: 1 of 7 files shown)


colinleach:
Scrap this. I'm not in best health (again) today, and failed to create a separate branch. One of those days when I seem to have forgotten how to code in any language, despite doing it for nearly 50 years...

PR #312, opened 2024-01-22
[Protein Translation] add new practice exercise

colinleach:
For week [02-13], so not _very_ urgent. Again, just the exercise with no approaches yet.

I had hoped to do something clever and vectorized with this exercise, but tripped over the endless validity-checking. This needs to be done in a specific order, so I ended up putting most things in a `for` loop. 

Maybe we could translate the whole sequence, then truncate to everything before the first "STOP" codon, then check for invalid elements? Probably still not elegant, and I've done enough molecular biology in my career to have a feel for how big real-world sequences can get - translating onward past a STOP isn't great.

### DIFF
--- a/config.json
+++ b/config.json
@@ -601,6 +601,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1
+      },
+      {
+        "slug": "protein-translation",
+        "name": "Protein Translation",
+        "uuid": "dc8bb817-1c79-476a-a666-bcad9455c66a",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
       }
     ]
   },
--- a/exercises/practice/protein-translation/.docs/instructions.md
+++ b/exercises/practice/protein-translation/.docs/instructions.md
@@ -1,0 +1,45 @@
+# Instructions
+
+Translate RNA sequences into proteins.
+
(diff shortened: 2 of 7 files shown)


jonmcalder (changes requested):
Thanks this looks good!

To be clear the only change I'm expecting here is uncommenting that test case, the rest below is just for consideration.

I'm never particularly happy to see a for loop in R, but as you say I think it may be called for here given the desire to process sequentially until a stop codon is reached. Maybe the alternative you outlined could be explored as another approach, or we can just wait and see what people come up with (I can't seem to think of anything right now). 

Another small consideration here is whether instead of naming the function "proteins" it could be "translate" or "translate_rna" or something similar?
Not really an R specific thing, but I do think it's good to encourage (and model) the principle of preferring verbs for function names to describe what they do. I may not always have done this in the past with the R track, but it's just something that occurred to me when reviewing this and I thought it worth raising (naming things is hard, and actually a massive and under appreciated aspect of good programming in my view)

I assume here that you used ""proteins" simply because that's what the method was called in the python implementation, and scanning a few of the other tracks I see that has been done quite a lot, while others have opted for more verb-like function names. So I'm not sticky on it at all, and it can probably be debated different ways - was just keen to raise it and hear your view.

jonmcalder:
Thanks! I like `translate()`.

jonmcalder (approved):

PR #315, opened 2024-01-30
Update INSTALLATION.md

rainerrodrigues:
I have added another bullet point for optional installation of Anaconda so that User don't have to install additional IDE for both the Python and R track.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -8,6 +8,11 @@ RStudio is a popular cross-platform integrated development environment (IDE) for
 
 Download and install the [current stable version of RStudio](https://www.rstudio.com/products/rstudio/download/). Or, alternatively, get the [preview version](https://www.rstudio.com/products/rstudio/download/preview/) of an upcoming release.
 
+## Alternatively Install Anaconda (Optional)
+Anaconda® Distribution is a Python/R data science distribution that contains conda, a package and environment manager, which helps users manage a collection of over 7,500+ open-source packages available to them.
+
+Download and install the [current stable version of RStudio](https://docs.anaconda.com/free/anaconda/install/).
+
 ## Install the R packages for running tests
 The test runner for the specs is the [`testthat`](https://github.com/hadley/testthat) library from Hadley Wickham, which is a popular choice for R package authors.
 


colinleach:
I'm not sure how we feel about Anaconda, so I'll leave @jonmcalder to comment on that. My own prejudice is that Anaconda brings in a lot of things, so I prefer to install Miniconda for the base environment and other packages as needed for each further environment (Python, R, Julia, whatever). Opinions no doubt differ!

Reading through the whole of this `installation.md` document, I was struck by the last paragraph that mentions `magrittr`. I think we should update this to mention `tidyverse` which was recently added to the test runner.

jonmcalder:
Hi @rainerrodrigues! Thanks for raising this.

And thanks @colinleach and @BethanyG for weighing in here.

I think @BethanyG has said it better than I could and I agree that many of the same considerations apply from the R perspective.

In my experience Anaconda (or other conda variants like Miniconda etc) are commonly encountered and useful when working in "scientific" computing contexts where somewhat heavy package / dependency / environment management can be required and sometimes as a way of facilitating interop between python and R , but as already pointed out neither python nor R are confined to scientific computing (just strongly affiliated) and even in this space different tradeoffs / preferences will apply for different users and use cases so Anaconda is not something I'd suggest or push strongly in the Exercism context or for beginners.

That being said however, I do concede that the current installation doc pushes RStudio strongly without mentioning other popular alternatives like VSCode, Emacs, Vim etc. Do you guys think it should be softer in that regard and suggest more alternatives? Or not mention a specific IDE at all?

Up until now my thinking has been that those who really want to use an IDE or editor they're already familiar with will explore that option naturally on their own either upfront or whenever suits them. But for someone who doesn't know where to start, I think being pointed in a single direction is better than being given a range of options and for starting out with R, I believe RStudio is the most obvious choice because of the way it ties together an editor with REPL, help docs, viewer / environment pane etc without needing to setup any plugins or config.

If anything else is to be added, my preference would be to mention the [R Extension](
https://code.visualstudio.com/docs/languages/r) for VSCode which is widely used by the R community (and also because many will already use VSCode as an IDE for other languages).

And yes @colinleach you're right that we should update the {magrittr} bit to mention the {tidyverse} instead now that it's supported by the rest runner.

@rainerrodrigues would you be willing to update your PR in line with what's agreed upon here? (assuming we haven't put you off)

BethanyG:
> That being said however, I do concede that the current installation doc pushes RStudio strongly without mentioning other popular alternatives like VSCode, Emacs, Vim etc. Do you guys think it should be softer in that regard and suggest more alternatives? Or not mention a specific IDE at all?

I think its true that many folx who go to learn R end up installing RStudio because it has been around a while, and Posit is so active in the r-related data science community.  It also gets recommended a lot for courses (or at least it did).  It is very easy to download and set up.  So some students may not be aware that it is even possible to install R and use a different IDE or editor.  So that might be argument enough for providing at least 1-2 alternatives.  Or not...

We had an analogous issue on the Python track where PyCharm has a lot of mindshare, as does JupyterLabs for data-sciency stuff.  And then there was VSCode eating the code-editor universe (_many online coding sites have or use the Embedded VS code editor - including codespaces here on Github_).

So for our docs, we wanted to make sure students were aware that they _did not_ need to buy and IDE (PyCharm or others), and that they also _did not_ have to install a bunch of the scientific python stack and use Jupyter - which wouldn't help them much at all with running tests or solving exercises.

So we ended up listing out a bunch of stuff, starting with VSCode.  I think it probably had the net effect of steering most students that direction, but at least they can scan down and try out others if they feel like it. 🤷🏽‍♀️ 

We do, however separate [installation](https://exercism.org/docs/tracks/python/installation) from [development tools](https://exercism.org/docs/tracks/python/tools).  And we point at the PSF installation method rather than any distro like Anaconda or homebrew.

colinleach:
I think this is superseded by #316 and can be closed.

jonmcalder:
Thanks again to all 3 of you for your input on this - especially @BethanyG for your insight.

PR #317, opened 2024-02-01
Pop count - Added practice exercise

Nerwosolek:
I've added pop-count exercise as @colinleach and @jonmcalder suggested.

Somehow during autogeneration config.json was changed in other parts also, I guess only for formatting issues.

### DIFF
--- a/config.json
+++ b/config.json
@@ -50,7 +50,9 @@
         "slug": "lasagna",
         "name": "Lasagna",
         "uuid": "6fd75bbd-8654-4d6a-80b5-29344ceb8a48",
-        "concepts": ["basics"],
+        "concepts": [
+          "basics"
+        ],
         "prerequisites": [],
         "status": "wip"
       },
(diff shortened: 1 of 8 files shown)


jonmcalder:
@Nerwosolek Thanks! Will review this when I get a gap.

@colinleach you'll probably beat me to it so feel free to review and provide feedback or merge if you're happy.

colinleach (approved):
Looks great, thank you for submitting.

I suggested a small change to `example.R`, using the same functions with native pipes rather than nested. This is a newer style that is now encouraged.

I'm puzzled by all the changes in the top-level `config.sys`. These look generally sensible and the sort of think that `configlet` might do automatically, but not something I've seen before. I've held off merging until @jonmcalder has had chance to look at this and give an opinion.

Alternative approach: @BethanyG, have you seen anything like this on the Python track?

jonmcalder:
I agree that the changes to config are just formatting (whitespace and ordering of fields), so presumably that's just a recent-ish improvement to configlet.

Happy to merge - thanks again @Nerwosolek and to @colinleach for reviewing!

jonmcalder (approved):

PR #319, opened 2024-02-01
:memo: Format exercise meta via configlet

jonmcalder:


colinleach (approved):
This seems uncontroversial.

PR #323, opened 2024-02-10
Loop over romans fix

Nerwosolek:
- changed config.json in approaches
- renamed `loop-over-digits` to `loop-over-romans` directory in approaches


colinleach (approved):
This looks good - thanks!

PR #326, opened 2024-03-08
[New practice exercise] Robot simulator

Nerwosolek:
As we talked I proposed the approach with S3 class called `"robot"` and one generic method and one class method.
Becuase it is hard to force the implementation in this case I tried to suggest the implementation via:

- Tests - one test if S3 class was defined and the others use class constructor. I cannot find a test for proper method dispatching, so `move` function could be also normal function.
- Stub, which suggests object constructor, generic method and class method.

In `example.R` I proposed how I imagine S3 implementation for this exercise. I wanted it to be simple, cause this exercise does not justify full blown S3 implementation as there is only one class we can think of.
WDYT? Can I improve it somehow?


colinleach:
I don't know why the linter is complaining about the stub file, but not the `example.R`. We'll need @jonmcalder to comment on that, and any other linter issues.

I'll provisionally approve this now, but merging will need to wait for Jon's comments.

colinleach (approved):
Approved provisionally, but linter issues need input from another reviewer.

Nerwosolek:
> Approved provisionally, but linter issues need input from another reviewer.

I addressed some of the linter complaints but there are some inner errors of linter itself.
Look in the forum. I could take a look at this in the other PR.

Nerwosolek:
> I don't know why the linter is complaining about the stub file, but not the `example.R`.

I suspect linter can check if `move.robot` function belongs to the class `robot` of S3 system, maybe via `methods()` function.

Nerwosolek:
@colinleach Could you look again? I solved the problem but the fact I had to use `# nolint` and it will be visible to student does not satisfy me.

exercises/practice/robot-simulator/.meta/example.R
```diff
@@ -0,0 +1,46 @@
...
+    coordinates = numeric(2),
+    direction = c("NORTH", "EAST", "SOUTH", "WEST")
+  ) {
+  stopifnot(is.numeric(coordinates))
+  stopifnot(length(coordinates) == 2)
+  direction <- match.arg(direction)
+  
+  structure(
+    list(
+      coordinates = as.integer(coordinates),
```
jonmcalder, line 11: Could your talk me through your thinking here?
It feels weird to me to define coordinates as numeric on line 2 and then convert to integer.

It's late for me here so I might be overlooking something, but I think you could just change this (line 11) to coordinates = coordinates.

An alternative would be to be stricter and expect an integer (and check with is.integer in line 5), though then one would need to pass in the coordinates as e.g. `c(0L, 0L)` and the tests would also all need to be updated to conform to this expectation.

jonmcalder (changes requested):

jonmcalder:
Thanks this looks good!
And thanks also for addressing the outdated linter setup.

I'm not too fussed about adding in the `#nolint` comment in the stub to avoid the linter complaining - I was going to suggest that at least as a stop gap solution for now.
I did also consider it might confuse students slightly or be off-putting but even if it's not immediately clear a quick Google should make it obvious why it's there and it doesn't necessarily need to stay as part of their solution.

I just have the one gripe regarding types - feels like we should probably be a little stricter and expect integers, otherwise if we're ok with numerics then the example constructor shouldn't convert the type when creating the object.

exercises/practice/robot-simulator/.meta/example.R
Nerwosolek: Above all I wanted to have default value for position as `(0, 0)`.
I started as you also suggested with `integer(2)` although I find it very unnatural for user to pass `c(1L, 1L)` so I went to numeric. 
Than I thought that such object could internally represent position in integer, helping user with the coordinated when hipothetically such robot object would be used according to requirements but accidentally passed non-integer values like `c(1.1, 3.5)`.

At the same time our tests does not force internal representation as `0 == 0L`

Nerwosolek:
> should prompt students to implement a move generic but since the tests don't actually require or test for that it could be done > either way which I think is ok so (pending the issue regarding numeric vs integer types) I'm happy with this implementation as is > if you and @colinleach are.

I'm OK with that of course :). 
Maybe it's a point for discussion during mentoring session about generics and dispatching.
I just wanted to introduce anything more than just a class (as that was already in triangle exercise). 
Maybe later there will be exercises which give opportunity to build even more sophisticated S3 system based on already acquired knowledge.

exercises/practice/robot-simulator/.meta/example.R
jonmcalder: Ok I figured you might have been thinking along those lines.


I agree that expecting `c(0L, 0L)` might be a bit disruptive / confusing for (some) students. I also wondered whether some students might wish to implement it with stricter integer types (which is arguably more correct) and can't do so because the tests won't pass with their numeric inputs.

But the tradeoffs you propose sound like they might be the best course of action.

Could I then suggest that we give this exercise beta status for now in the config to encourage feedback and communicate that it might still change slightly in future?

config.json
```diff
@@ -639,6 +639,14 @@
...
           "math"
         ]
+      },
+      {
+        "slug": "robot-simulator",
+        "name": "Robot Simulator",
+        "uuid": "67e94669-bea1-43f4-9c2f-90884676f568",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1
```
jonmcalder, line 649: Can we add `"status": "beta"` here?

jonmcalder (changes requested):

exercises/practice/robot-simulator/.meta/example.R
Nerwosolek: Sure. I don't know what are the features of "beta" exercise but if you find it useful I'll just drop the change.

jonmcalder (approved):

jonmcalder:
Good work and thanks again!

config.json
depial, line 1: Not sure if I'm a bit late here, but was difficulty meant to be left at 1? Just for context: this exercise shows up as "Medium" rather than "Easy" in the Julia track.

Nerwosolek: A little bit late but not a problem we can always drop a new change if we agree on something :)
Please notice that on C# and Python track it is "Easy" exercise at the same time. I find it hard to evaluate difficulty, especially without criteria. When I was starting solving exercises on Exercism I thought that difficulty is more connected with the alogrithmic hardness of a problem not a sofistication of language features used.
For me this is basic introduction to OOP in any language having any kind of object orientation.

depial: I did think about how the exercise is straightforward algorithmically, and the class structure used here isn't terribly complicated, but I tend to think of difficulty more in terms of where in a learning journey a learner would normally pick something up.

That being said I can see how language sophistication in this case can be more of an issue in Julia, since its not object oriented, and I've found dealing with constructs and the type system is not necessarily easy, and therefore medium :D

Anyway, I just thought I'd ask since I noticed that configlet seems to assign everything difficulty 1 by default (at least in the two exercises I've used it for), and I don't think I would have found this to be easy if I were fairly new to either R or OOP (the reference is helpful though!). Feel free to leave it or change it as you see fit, since I'm just a casual user of R and I'm sure you're more familiar with it than I am :)

In any event, thanks for contributing the exercise!

Nerwosolek: Perhaps if we had concepts tree this exercise would be placed further in the journey through the language.

depial: Yeah, I could see it coming later, but I'm not sure if I have a biased view since all of my limited 'formalish' R instruction has been in the context of data science, and I don't remember seeing OOP introduced in those courses.

I took a quick look at the reference provided for the exercise (Advanced R - Hadley Wickham), and OOP comes about halfway through, so it also seems to place it later in the learning journey.

In the end, it's a fairly minor detail and I imagine it depends on the target audience. I would say it's definitely easy for people with a moderately sophisticated programming background, but likely more difficult for people starting out.

PR #333, opened 2024-04-18
Fix phone number tests

Nerwosolek:
This PR in phone-number exercise:

- Synced via configlet tests.toml with canonical
- Deleted two non-canonical tests, added 5 missing canonical and a few reimplemented tests
- changed example.R to passing new tests.


colinleach (approved):
Thanks for this. It seems to bring things in sync with problem-specifications.

PR #345, opened 2024-09-11
Add script to add practice exercise

ErikSchierboom:
I'm planning on maybe adding a couple of exercises, and this script helps (newer tracks get this for free: https://github.com/exercism/generic-track/blob/main/bin/add-practice-exercise)

### DIFF
--- a/bin/add-practice-exercise
+++ b/bin/add-practice-exercise
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+
+# Synopsis:
+# Scaffold the files for a new practice exercise.
+# After creating the exercise, follow the instructions in the output.
+
+# Example:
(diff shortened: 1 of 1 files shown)


jonmcalder (approved):
Thanks Erik! Have only checked / scanned this quickly from my phone for now so I'll want to actually test it out later but I approve nevertheless. Is the idea to keep this generic or to (optionally) tailor it further to each track?

ErikSchierboom:
> Is the idea to keep this generic or to (optionally) tailor it further to each track?

It's a generic starting point, but each track should tweak it as they desire.

ErikSchierboom:
I'll wait with merging until you've tested it.

jonmcalder:
Thanks I have since run it and all looked good so I think we can merge.

PR #346, opened 2024-09-11
series: add exercise

ErikSchierboom:


colinleach:
The R track has a particularly fierce linter. We all get caught by this the first few times!

ErikSchierboom:
I've updated the PR

config.json
```diff
@@ -648,6 +648,14 @@
...
         "difficulty": 1,
         "status": "beta"
+      },
+      {
+        "slug": "series",
+        "name": "Series",
+        "uuid": "2ef6f027-a27c-4b00-bf1e-2969baf04a36",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 3
```
jonmcalder, line 658:
```suggestion
        "difficulty": 3,
        "topics": [
          "strings"
        ]
```

ErikSchierboom: I'm fine adding this, but you do know that `topics` aren't used by the website, right?

jonmcalder: Haha, it did cross my mind that if you hadn't included it, it was quite likely because it's not really needed or being used anywhere, but was more efficient for me to suggest it and wait and see how you responded than for me to try and check if or how it might be used later. If there's no value in adding it (now or for later) then obviously fine with leaving it out.

PR #347, opened 2024-09-16
Add `game-of-life` exercise

ErikSchierboom:
I think this should work, but I can't seem to get the tests passing.

Also, let me know if using a matrix is a good idea here, or if I should use something else

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
-      
+
       - uses: r-lib/actions/setup-r@v2
         with:
           r-version: 4.3.1
@@ -40,18 +40,18 @@ jobs:
 
     strategy:
       matrix:
-        R: [ '4.3.1' ]
+        R: ["4.3.1"]
 
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
-      
+
       - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.R }}
 
       - name: Install project dependencies
         run: |
-          install.packages(c("testthat", "jsonlite", "lintr"))
+          install.packages(c("testthat", "jsonlite", "lintr", "dplyr"))
         shell: Rscript {0}
 
       - name: Run exercism/r ci (runs tests) for all exercises
--- a/config.json
+++ b/config.json
@@ -656,6 +656,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 3
+      },
+      {
+        "slug": "game-of-life",
+        "name": "Conway's Game of Life",
+        "uuid": "5ba723f2-dd9e-452a-b5f3-cfe19a2650b5",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4
       }
     ]
   },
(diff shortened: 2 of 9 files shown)


colinleach:
That made me think for a while!

It seems that the `intersect()` on line 6 isn't doing quite what we expect. On debugging, `neighbor_coords` is an empty list, which is obviously wrong (or at least unintended).

Now that we have `tidyverse` in the test runner, it is possible to add `library(dplyr)` at the top of the file, then replace `intersect` with `inner_join`. All tests passed when I tried this locally.

colinleach:
I suspect @jonmcalder would have a better solution, if he is around.

colinleach:
Also, using a matrix looks fine to me. Maybe my judgement is warped by long experience with Matlab and NumPy, but I don't see why it shouldn't also be idiomatic R. This data type has been in the base language for a long time.

ErikSchierboom:
Thanks @colinleach! I'll wait for @jonmcalder to see if he agrees too, and then we I'll continue fixing things.

jonmcalder:
Yes I agree that it makes sense to use a matrix implementation for this.

Colin is correct that you can't use `intersect()` in that context because it's meant to work on vectors (and `expand.grid()` produces a data frame). Using either `dplyr::intersect()` or `dplyr::inner_join()` instead would work here.

I don't have a different approach to suggest right now (and can't give more time to this) but I think there are some interesting approaches [here](https://gist.github.com/oliviergimenez/8ee33aede25925e84bda67342eb49b67) if you want to explore a little more. 

The implementation and tests are the main thing and in my view the example solution only really needs to demonstrate a possible way to solve the exercise (ideally as elegantly and idiomatically as possible but that can be subjective). So I'm sure I'll be happy with whatever you guys decide here as long as you both are happy.

ErikSchierboom:
Minor suggestion: maybe the linting should be moved to its own, separate GHA workflow. That will allow the contributor to verify that the tests are passing without having to satisfy the linter.

ErikSchierboom:
> ideally as elegantly and idiomatically as possible but that can be subjective

I'll happily let others submit a PR to improve upon my solution :)

jonmcalder:
> Minor suggestion: maybe the linting should be moved to its own, separate GHA workflow. That will allow the contributor to verify that the tests are passing without having to satisfy the linter.

Yes probably a good suggestion. Out of interest have did you consider using {styler} and/or {precommit} to help avoid linting issues as outlined in the (now very outdated I'm sure) ReadMe?

ErikSchierboom:
I hadn't done that. Probably should have. Sorry

ErikSchierboom:
@jonmcalder I've fixed CI

jonmcalder (approved):
Cool - I'm happy with this implementation.

Still hoping someone will put up their hand to improve on Erik's example solution though - not sure if it's the nested functions or the mixing of operations on matrices and data frames with apply / expand.grid / dplyr::inner_join but it's not easy to read and seems like a bit of a code smell to me.

(no offence of course @ErikSchierboom - you're a polyglot so you're bound to approach things differently and code can be subjective - thanks for contributing this exercise!)

ErikSchierboom:
> no offence of course @ErikSchierboom

None taken! There must be someone that can do this a ton better.

ErikSchierboom:
I'm merging this to have it be available, and then the updated example solution can be PR'ed later.

colinleach:
I hadn't previously thought about the impact on CI, after adding `tidyverse` to the test runner. I suspect that at some point we will want more than just `dplyr` available. I'm thinking particularly of `stringr`: a pretty major improvement on string handling in base R.

I understand the bias towards keeping CI lean, but what are the practical implications of adding more packages? Slower? More expensive?

PR #350, opened 2024-10-01
knapsack: add exercise

colinleach:
The linter is going to complain. Too bad, it's a long URL inside a comment, so live with it.

More seriously, I'm not thrilled with the `example.R`. I had hoped to vectorize this, but each cell update needs access to the cells above and to the left. Hence needing both matrix indices, hence nested loops. If you have a better idea, please change the file and add yourself as an author.

I've put difficulty 5, copying Python, Julia, C#.

### DIFF
--- a/exercises/practice/knapsack/.meta/example.R
+++ b/exercises/practice/knapsack/.meta/example.R
@@ -1,5 +1,4 @@
 maximum_value <- function(max_wt, items) {
-  # This URL is > 80 chars, ignore whining from the linter:
   # nolint start
   # https://en.wikipedia.org/wiki/Knapsack_problem#Dynamic_programming_in-advance_algorithm
   # nolint end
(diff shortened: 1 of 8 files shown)


exercises/practice/knapsack/.meta/example.R
```diff
@@ -0,0 +1,28 @@
+maximum_value <- function(max_wt, items) {
+  # This URL is > 80 chars, ignore whining from the linter:
+  # https://en.wikipedia.org/wiki/Knapsack_problem#Dynamic_programming_in-advance_algorithm
```
ErikSchierboom, line 3:
```suggestion
  # nolint start
  # https://en.wikipedia.org/wiki/Knapsack_problem#Dynamic_programming_in-advance_algorithm
  # nolint end
```

ErikSchierboom: This fixes the linting problem

exercises/practice/knapsack/knapsack.R
```diff
@@ -0,0 +1,3 @@
+maximum_value <- function(max_wt, items) {
```
ErikSchierboom, line 1: Maybe make this slightly more verbose?

```suggestion
maximum_value <- function(max_weight, items) {
```

colinleach:
Thanks for the linter advice. I failed to RTFM!

ErikSchierboom (approved):

PR #355, opened 2024-10-11
Add `rail-fence-cipher` exercise

ErikSchierboom:


exercises/practice/rail-fence-cipher/.meta/example.R
ErikSchierboom, line 1: I'm obviously biased, but I think this example implementation is one of the nicest I've seen. R is so powerful!

colinleach: I'm not going to disagree!

colinleach (approved):

PR #358, opened 2024-10-11
flatten-array: fix nesting

ErikSchierboom:
I erronously used `c()` for nesting, but that won't actually create nesting. Switched to `list()` Sorry!


colinleach:
Ouch, it's almost as though the reviewer wasn't doing his job...

colinleach (approved):

colinleach:
I'm assuming no students have done this new exercise, so there's nothing to re-test.

PR #360, opened 2024-10-12
Add `matrix` exercise

ErikSchierboom:


colinleach (approved):
Linters are valuable, but I find this one a bit over-pedantic for my tastes.

colinleach:
It took me a while to work out why you need both `unlist` and `as.vector`. It really tries hard to hold on to the names otherwise.

Use of `unname` instead of `as.vector` does the same thing, and may be a bit more clear in its intention?

ErikSchierboom:
Fair point! i've updated

PR #366, opened 2024-10-15
Add `strain` exercise

ErikSchierboom:


colinleach:
The exemplar is longer than it needs to be, though it took me a while to see how to improve it (too much Julia going on in my old, tired head recently). Suggested alternative:

```r
keep <- function(input, fun) {
  if (is.null(input)) return(c())
  input[sapply(input, fun)]
}
```

I kind of feel there should be a one-liner using `ifelse`, but that struggles with a NULL on one branch and a vector on the other.

colinleach:
After writing my previous comment, I realized there is no need to test for null input. Only the `input[sapply(input, fun)]` is needed, so it is a one-liner.

ErikSchierboom:
@colinleach That is really cool, but it goed against the instructions:

> Keep your hands off that filter/reject/whatchamacallit functionality provided by your standard library!
Solve this one yourself using other basic tools instead.

colinleach (approved):

PR #365, opened 2024-10-15
Add `run-length-encoding` exercise

ErikSchierboom:

PR #368, opened 2024-10-15
CI: add tidyverse

colinleach:
I tried adding `tidyverse` as a full package, but got alarmed when GH took 11m36s to install the dependencies (lots of C++17 compilation).

Thinking about it some more, we definitely don't need `ggplot2` (a big plotting package), and probably don't need `tidyr` or `forcats` (mainly about data cleaning). Adding `"dplyr", "stringr", "readr", "purrr", "tibble"` brought installation time down to 6m20s. I assume GH will cache the results somewhere?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install project dependencies
         run: |
-          install.packages(c("testthat", "jsonlite", "lintr", "tidyverse", "dplyr"))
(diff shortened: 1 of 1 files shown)


colinleach:
For anyone wondering, this is prompted by PR #365 and problems Erik hit in that case.

ErikSchierboom:
> I assume GH will cache the results somewhere?

Not automatically, no. But let's see how bad it is. Maybe the R action _does_ add some caching. If not, we can add it ourselves.

PR #369, opened 2024-10-16
Use GHA workflow to install dependencies

ErikSchierboom:
@colinleach According to the docs, using [this workflow](https://github.com/r-lib/actions/tree/v2/setup-r-dependencies) _will_ cache things, which should greatly speed up our builds (after the initial run).


ErikSchierboom:
This shaves off a couple of minutes each run.

PR #374, opened 2024-10-17
Use more specific test functions

ErikSchierboom:
See https://www.rdocumentation.org/packages/tinytest/versions/1.4.1


colinleach (approved):

colinleach:
I think `testthat` has evolved quite a lot since many of these tests were written, and even newer ones just copied the old style.

PR #386, opened 2025-01-22
Add `change` exercise

ErikSchierboom:


exercises/practice/change/.meta/example.R
```diff
@@ -0,0 +1,30 @@
+find_fewest_coins <- function(coins, target) {
+  if (target == 0) {
+    return(c())
+  }
+
+  stopifnot(target > 0)
+  stopifnot(min(coins) <= target)
+
+  min_amount_coins <- rep(Inf, target + 1)
```
ErikSchierboom, line 9: I struggled a bit with how to model the "the value is not yet there at that index" placeholder. I'm open to suggestions how to improve

colinleach: Conventionally, this would use [`NA`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/NA). Line 9 then becomes

```R
  min_amount_coins <- rep(NA, target + 1)
```
and line 27 becomes

```R
  stopifnot(!is.na(min_amount_coins[[target + 1]]))
```

The tests still pass when I tested this locally. Is that the sort of thing you were asking about?

colinleach (approved):
I'll approve but not merge for now. Make changes if you wish, then merge when you're ready.

ErikSchierboom:
Yeah, I'll to make it clearer using `NA`.

PR #390, opened 2025-02-23
[complex-numbers] Added new concept

colinleach:
As requested, though we still don't have a Concept Exercise to pair with it. I'll mark it draft until we can get an `introduction.md`.

Meanwhile, I had a look through some of the concepts I PR'd back in 2023, and was shocked how bad they are. That was before I had the highly educational experience of working with Bethany on the Python track (she's a very careful teacher). I've changed their status to draft, and I'll try to rewrite them sometime when I have the time, energy and functional brainpower simultaneously. I haven't looked at the old Concept Exercises yet, I _hope_ they are in better shape.


concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,239 @@
...
+`Complex numbers` are not complicated.
+They just need a less alarming name.
+
+They are so useful, especially in engineering and science, that R includes [complex numbers][complex] as standard 
+[atomic types][atomic] alongside integer, double, logical, character and raw.
+
+## Basics
+
+A `complex` number in R is a single value.
+In use, however, it can be thought of as a pair of numbers: usually but not always floating-point.
```
Nerwosolek, line 13: What do you mean here? I find it confusing as `?complex` states that: "Internally, complex numbers are stored as a pair of double precision numbers, either or both of which can be NaN(including NA, see NA_complex_ and above) or plus or minus infinity."

concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,239 @@
...
+
+> Arg(z1)
+[1] 0.9272952
+```
+A partial explanation, for the mathematically-minded _(again, feel free to skip)_:
+
+- The `(real, imag)` representation of `z1` in effect uses Cartesian coordinates on the complex plane.
+- The same complex number can be represented in `(r, θ)` notation, using polar coordinates.
+- Here, `r` and `θ` are given by `abs(z1)` and `Arg(z1)` respectively.
+
```
Nerwosolek, line 173: What do you think about adding some example how can we rotate 2D vectors in R easily.
Like in here we rotate vector (1,2) by 90 degrees leftwise:
```
x <- 1
y <- 2
angle <- pi/2
rot_z <- complex(modulus = 1, argument =  angle)
rot_z * (x + 1i*y)
```

Nerwosolek (approved):
I had some suggestions, if you don't find them interesting just merge.
If you don't have time and would like to implement some of them, let me know if I can help.

concepts/complex-numbers/about.md
colinleach: We included a lot on this in the Julia syllabus, which is aimed at an engineering and physics audience. I'm hesitant about getting into this detail in the R syllabus, aimed more at data science, statistics, epidemiology, social sciences, etc.

concepts/complex-numbers/about.md
colinleach: Truncated line 13

colinleach:
Time to get this reactivated, now we have a live syllabus to include it in.

I've:
- Made most but not all of the suggested changes
- Updated references to current standards
- Updated code formatting to current standards
- Fixed (I hope!) the merge conflicts in `config.json`
- Cleaned up some Markdown and JSON formatting

@depial, could you cast an eye over this briefly, and see if it fits with the rest of the syllabus?

I'm not sure what to do about an exercise. I think `tracking-turntable` is a good fit for Julia, not so much the likely audience for R (as I said in a previous comment, vector rotations aren't a big use-case for R).

Maybe something with a board game? For exercises like `flower-field`, treating the board as a complex plane can be an easy way to get neighboring squares (Re(z) and Im(z) each ±1).

colinleach:
I must also remember to copy the About into the Introduction before merging.

concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,242 @@
...
+```r
+r <- 2
+im <- 3
+
+r + im * 1i
+#> [1] 2+3i
+
+complex(real = r, imaginary = im)
+#> [1] 2+3i
+```
```
depial, line 41: I think it would be good to expand on this just a bit. Either by mentioning that the arguments are being handled as kwargs in this example (and therefore `real` and `imaginary` are necessary), or by explaining how to use the function's arguments positionally.

For example: `complex(real = r, imaginary = im) == complex(1, r, im)` evaluates to true since the first positional argument is the length of the output vector. I just think it would be good to clarify that `complex(real = r, imaginary = im) != complex(r, im)`, since that's an easy, and erroneous, logical leap to make.

I have to admit, that is one of the wackier ideas I've seen in a while. I'm unsure when someone would prefer having a repetition argument as the leading argument in a complex number constructor, but I'd imagine it probably has something to do with piping? That's R for you...

concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,242 @@
...
+#> [1] 3.752771+0.256554i
+```
+
+There are several functions, in addition to `Re()` and `Im()`, with particular relevance for complex numbers.
+Please skip over these if they make no sense to you!
+
+- `Conj()` simply flips the sign of the imaginary part of a complex number (_from + to - or vice-versa_).
+  - Because of the way complex multiplication works, this is more useful than you might think.
+- `abs(<complex number>)` is guaranteed to return a real number with no imaginary part and `Mod(<complex number>)` 
+  is a synonym.
```
depial: The mathematical description here is a bit vague and I don't think it would hurt to add a bit more specificity while foreshadowing your further example below. Something like:

> `abs(<complex number>)` is guaranteed to return a real number with no imaginary part that can be thought of as its "magnitude", and `Mod(<complex number>)` is a synonym.

concepts/complex-numbers/about.md
```diff
@@ -0,0 +1,242 @@
...
+    - These make no sense in classical physics and can only be explained (and designed) by quantum mechanics (QM).
+    - In QM, everything is complex-valued by definition. (_it's waveforms all the way down_)
+
+2. __The Fast Fourier Transform algorithm__.
+    - FFT is an application of complex numbers, and it is in _everything_ connected to sound transmission, audio processing, photos, and video.
+
+    - MP3 and other audio formats use FFT for compression, ensuring more audio can fit within a smaller storage space.
+    - JPEG compression and MP4 video, among many other image and video formats, also use FFT for compression.
+
+    - FFT is also deployed in the digital filters that allow cellphone towers to separate your personal cell signal from everyone else's.
```
depial: Are the extra spaces here intentional?

depial (approved):
This looks good to me! Just a couple of minor suggestions. I'm sure it'll be satisfying to get this out of the open PR list :D

depial:
> Maybe something with a board game? For exercises like flower-field, treating the board as a complex plane can be an easy way to get neighboring squares (Re(z) and Im(z) each ±1).

I can see this as a good way to incorporate the idea without diving into the math too much. From my initial feeling, it might be hard to get anything more than addition, subtraction and maybe an absolute value in there, but it seems doable, since actually constructing the numbers could do some of the lifting too.

Edit: Actually, I could see doing something with streets on a grid, so we could incorporate L1 and L2 norms more naturally.

concepts/complex-numbers/about.md
depial: On second thought, "can be thought of as" could probably just be "is"

concepts/complex-numbers/about.md
colinleach: No, just some copy-paste weirdness

concepts/complex-numbers/about.md
colinleach: I went with `(its "magnitude")` for brevity

concepts/complex-numbers/about.md
colinleach: Shit, I didn't even know about the weird-first-arg thing. I've added a comment: must include the `real` and `imaginary` keywords.

It probably made sense to the people designing S back in the 1970s, and isn't important enough for the Tidyverse team to fix with a modern implementation.

depial:
I think I've got enough of an idea for an exercise, involving taxis, that I could put it together if you'd like.

PR #396, opened 2025-06-17
[r GH Workflow Tooling]: Update lint.yml to ubuntu-24.04

BethanyG:
Upgrade to ubuntu-24.04.


colinleach:
see #395 for comments

PR #399, opened 2025-06-29
Adding exercise "palindrome-products" to track "r"

frequenter-at-exercism:
Didn't implement test `[12e73aac-d7ee-4877-b8aa-2aa3dcdb9f8a]`, as the schema I used to assumes the `$smallest` and `$largest` are both calculated with every call to `palindrome_products`.

### DIFF
--- a/config.json
+++ b/config.json
@@ -939,6 +939,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 5
+      },
+      {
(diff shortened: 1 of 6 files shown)


colinleach (approved):

colinleach:
Looks good, thank you!

Difficulty 6 seems about right. Python and Julia chose 4, C# is 6, and to the student they are all Medium.

PR #401, opened 2025-06-30
Add `variable-length-quantity` exercise

frequenter-at-exercism:
All tests implemented, all tests passed, all lints trapped.


colinleach:
Thank you! You are getting good at this, and we're very happy to see that. The R track could certainly benefit from some love and attention, these days.

PR #415, opened 2026-02-02
Update config.json

depial:
The practice exercises were not ordered by difficulty. This is addressed here. 

No particular order has been followed outside of the difficulty level (and keeping `Hello World!` first). 

I didn't look too carefully at which exercises received what rating, but it might be good to have a second look through the difficulty 1 group since that is the default assignment given by configlet, and so it's possible to create a more difficult exercise and miss changing that bit (e.g. is `roman-numerals` a 1?)

### DIFF
--- a/config.json
+++ b/config.json
@@ -265,7 +265,7 @@
         "uuid": "366d83ae-a069-457e-ace9-f79417ed311f",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 1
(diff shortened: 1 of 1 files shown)


depial:
The Configlet check got cancelled because:

> The job was not acquired by Runner of type hosted even after multiple attempts

I've never seen that before...

colinleach (approved):

PR #416, opened 2026-02-03
Add hints to annalyns-infiltration

depial:
Also updates `introduction.md`.

Notes:

- Ported `hints.md` from Julia track. 
- Again used `rdocumentation.org`.
- Added an extra code example to `introduction.md` to complement the last one. This can be removed if desired.

### DIFF
--- a/exercises/concept/annalyns-infiltration/.docs/hints.md
+++ b/exercises/concept/annalyns-infiltration/.docs/hints.md
@@ -1,1 +1,26 @@
 # Hints
+
+## General
+
+- There are three [boolean operators][boolean-operators] to work with scalar boolean values.
+- Multiple operators can be combined in a single expression.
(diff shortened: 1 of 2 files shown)


depial:
As an aside, have you considered putting `Vectors` second in the syllabus? It seems like, since R is inherently broadcast/vectorized, it might be appropriate to introduce them as soon as possible.

I just thought that then the explanation of the vector boolean operators could be mentioned at this stage (unless there was a later time for that to happen).

colinleach (approved):

colinleach:
> have you considered putting Vectors second in the syllabus?

I hadn't thought about it before, but now I am. Let me look into what's involved: I don't _think_ Elyse's Enchantments needs conditionals, but I need to check.

PR #424, opened 2026-02-06
Update Booleans concept

colinleach:
The vectorized `&` and `|` are now described in the the About, referred to obliquely in the Introduction. This is a bit of a balancing act, so please let me know if it works or not.


depial (approved):
Looking good here too!

The one thing that I have an itch to explain is the "order of operations" for logical operators and how they map `(!, &, |) --> (^, *, +)`. I'm not sure if that is out of scope or not though, but it could potentially be included with a single sentence if you think it could be useful.

PR #425, opened 2026-02-06
Update Conditionals concept

colinleach:
Big difference here between About (long section on `ifelse` and `if_else`) and the Introduction (same brief version as `cars-assemble`, except I changed one word). Is this a decent compromise?


depial (approved):
Two minor suggestions and a question.

I think the long `about.md` is fine. It's a good opportunity to introduce the availability of `tidyverse` and I think all the information in that section is relevant.

PR #426, opened 2026-02-06
Update Vectors concept

colinleach:
It's probably best to think of this as Part 1 of a Vectors series. I'll try to get Vector Filtering updated today, so you can see them together and we can decide if the content is split the right way.


concepts/vectors/about.md
```diff
@@ -117,63 +136,98 @@ w <- c(2, v[2:3], 11) # c(2, 7, 10, 11)
...
+v >= 6
+#> [1] FALSE  TRUE  TRUE
+all(v > 6)
+#> [1] FALSE    # not all elements match this condition
+any(v > 6)
+#> [1] TRUE     # at least one element matches
+```
+
+For more complicated logical expressions, it is important to use the `&` and `|` operators, which work on vectors.
+The double-character `&&` and `||` are only intended for comparing single values.
```
depial: A back reference to the conversation in `booleans` seems appropriate here.

colinleach:
> how vectors flatten

I added line 26

PR #427, opened 2026-02-06
Vector Filtering concept

colinleach:
First of the new ones!

This is pretty much Part 2 of Vectors.

As previously noted, we have a choice of exercises to go with it.


concepts/vector-filtering/about.md
```diff
@@ -0,0 +1,128 @@
...
+v[bools]
+#> [1] 2 3
+```
+
+It is a small step from there to generating the boolean vector with a conditional expression:
+
+```R
+v[v >= 2]
+#> [1] 2 3
+```
```
depial: Maybe we could split this up as an explicit reminder of what's going on:

```R
bools <- v >= 2
bools
#> [1] FALSE  TRUE  TRUE

v[bools]
#> [1] 2 3
```
However, I'm not against leaving as is either.

concepts/vector-filtering/introduction.md
```diff
@@ -0,0 +1,61 @@
+# Introduction
```
depial: Reminder about possible propagation.

depial (approved):
Just a couple small suggestions here. 

I was a bit hesitant to suggest adding more examples causing bloat because of the length of the `Vectors` Concept `about.md`, but this is quite a bit shorter so I feel more comfortable with the idea.

concepts/vector-filtering/introduction.md
colinleach: I also realized that the code samples were wrongly formatted.

colinleach:
+1 for `elyses-analytical-enchantments`. I see that `bird-watcher` has a comment in the exemplar describing my own solution as "clunky"!

There is probably a function somewhere in Tidyverse to improve it, but this is a very early Concept and we don't want to complicate it.

PR #428, opened 2026-02-06
Update config.json

depial:
Deprecate `Vehicle Purchase` practice\concept exercise.


depial:
As an aside, this made me wonder if concept exercises can have a difficulty field too? I don't remember ever using one on the Julia track.

PR #430, opened 2026-02-07
Vector Functions concept

colinleach:
I'll submit this as draft for now, because I'm not sure what to do with it.

I'm wondering if we could turn Bird Watcher into a suitable exercise, with a few modifications?

### DIFF
--- a/concepts/vector-functions/about.md
+++ b/concepts/vector-functions/about.md
@@ -95,6 +95,15 @@ pmax(v, w)
 
 This function and others like it also accept an arbitrary number of input vectors, not just two.
 
+The above example uses vectors of equal length.
+As discussed in the [Vector Filtering][concept-vector-filtering] Concept, R uses recycling to extend vectors which are somehow "too short".
+To repeat: _use this with care, and preferably only with a length-1 vector._
+
+```R
+pmax(v, 5)
+#> [1] 5 7 9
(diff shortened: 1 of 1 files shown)


concepts/vector-functions/about.md
```diff
@@ -0,0 +1,101 @@
...
+For example, consider the pairwise-max function [`pmax()`][ref-pmax]:
+
+```R
+v
+#> [1] 2 7 9
+w
+#> [1] 3 1 5
+pmax(v, w)
+#> [1] 3 7 9 # max of each pairwise comparison
+```
```
depial: If you want a non-numerical example, `paste` or `paste0` could work:

```R
part.1 <- c('Hel','Wo')
part.2 <- c('lo','rld!')
hi <- paste0(part.1, part.2)
hi
#> 'Hello'⋅'World!'

length(hi)
#> 2
```

colinleach: Something to think about!

 I've mostly steered away from strings, because they're not a prereq for this concept. Also because I haven't got far with writing the Strings concept: all the `paste` stuff is now largely replaced by `stringr`.

colinleach:
Working on `bird-watcher`, I needed a descending sort and realized I didn't mention it here.

PR #434, opened 2026-02-09
update docs

depial:
I just completed this online and caught a couple of typos. Also, I cut down the exercise `introduction.md` and perpetuated a change made in `about.md`.

One other thing to note. I'm starting to understand what you meant by nothingness causing issues with testing. Here, `NULL` is expected in the tests, which I would think we should avoid since this hasn't been introduced. I think it's much more natural to accept the `numeric(0)` that will come from operating on the vectors with obvious functionality.

If you'd like to make a change to that effect, I can include it in this PR.

EDIT: Actually, we can make a non-breaking change that just accepts something that has length 0, since `NULL`, `numeric(0)`, `integer(0)` etc all have length 0.

Instead of:
```R
test_that("remove the only card from the top", {
  stack <- c(1)
  expect_equal(remove_item_from_top(stack), c())
})
```
We could do:
```R
test_that("remove the only card from the top", {
  stack <- c(1)
  expect_equal(length(remove_item_from_top(stack)), 0)
})
```

### DIFF
--- a/exercises/concept/elyses-enchantments/.meta/exemplar.R
+++ b/exercises/concept/elyses-enchantments/.meta/exemplar.R
@@ -20,11 +20,7 @@ check_size_of_stack <- function(stack, stack_size) {
 }
 
 remove_item_from_top <- function(stack) {
-  len <- length(stack)
-  if (len > 1) {
-    return(stack[-len])
-  }
-  NULL
+  stack[-length(stack)]
 }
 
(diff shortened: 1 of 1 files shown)


depial:
> I suspect students at this stage will interpret `c()` (wrongly) as an empty vector.

Also, the failure feedback in the editor mentions `NULL` (not `c()`), so that's what the students end up seeing.

PR #436, opened 2026-02-10
Names attribute

colinleach:
Another one I'll park here as draft, because I'm not sure what to do with it. It's tentatively a prereq for Lists, but if necessary we could add it in at the top of that concept. Though it could then get a bit long.

I'll try to PR Lists later today, so it's clearer how they relate. It's setting the scene for dataframes (and Tibbles), where column names get _really_ important for navigation.


depial (approved):

depial:
Yeah, I was imagining something like `named-resistor-color` which has more and specific testing for named vectors. The current exercise might not even qualify to practice `names` since a more straightforward solution is with `which`:

```R
colors <- c("black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white")
color_code <- function(color) which(colors == color) - 1
```

Of course, I'm unsure if this is beating a dead horse, because there are already a few resistor themed exercises, but I thought it had a good amount of potential. I could try to throw together an example if you think it's an okay idea.

concepts/names-attribute/about.md
```diff
@@ -0,0 +1,95 @@
...
+```R
+> months[months == 30]
+Apr Jun Sep Nov 
+ 30  30  30  30 
+
+> sum(months)
+[1] 365  # no meaningful names possible
+```
+
+## More advanced topics
```
depial: Just a thought, now that I've thought more about it... It could be a useful "advanced topic" to talk about performance. I'm not sure if what I found is accurate, but it looks like indexing by name is a linear operation (rather than being stored as something akin to a hash map).

depial: I'm not sure about a solid link (as my searching might be failing me again), but I'm feeling more confident that it is a linear search to match, which would explain why having identical names will return only the first instance (and indeed why identical names are even possible):

```R
colors <- c("black", "brown", "black")
bands <- 1:3
names(bands) <- colors
bands["black"]

# => 1
```

If you're feeling adventurous and read C, here's the relevant [source code](https://github.com/wch/r-source/blob/trunk/src/main/subset.c), or a [StackOverflow](https://stackoverflow.com/questions/10499429/indexing-redundantly-named-vector-in-r?) conversation which mentions the source code (in `EDIT2` of the top answer)

In the end, `names` are just an attribute, so I don't think it changes the underlying data structure.

colinleach: > If you're feeling adventurous and read C

Not today. I seem to have eaten (or drunk) something which has left ... _side effects._ Better days will come!

PR #440, opened 2026-02-12
Loops concept

colinleach:
Pairs with `mixed-juices` and follows `switch`.

I think this is just old, traditional stuff, with nothing to add from Tidyverse.

### DIFF
--- a/concepts/loops/links.json
+++ b/concepts/loops/links.json
@@ -3,5 +3,4 @@
       "url": "https://intro2r.com/loops.html",
       "description": "An Introduction to R: Loops"
     }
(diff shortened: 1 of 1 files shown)


concepts/loops/introduction.md
```diff
@@ -0,0 +1,102 @@
+# Introduction
+
+There are basically two types of loops:
+
+1. Loop until a condition is satisfied.
+2. Loop over the elements in a collection.
+
+Both are possible in R], though the first may be more common.
```
depial: There's a stray bracket here.

PR #446, opened 2026-02-15
Add concept exercise named-resistor-color

depial:
This is meant to accompany the `names-attribute` concept. It is based on the `resistor-color` trio of exercises but is meant to push students down a `named vector` approach. To this end, the "practices" and "prerequisites" for the practice exercises `resistor-color` and `resistor-color-duo` have been set to `vector-filtering`, which is a prerequisite for this concept.

Notes:
- Tests have been grouped into single testsets. This can be changed to one test per testset for clarity if needed.
- Expected of student: create named vector, use `unname`, attempt named indexing.

### DIFF
--- a/exercises/concept/named-resistor-color/.docs/hints.md
+++ b/exercises/concept/named-resistor-color/.docs/hints.md
@@ -1,5 +1,7 @@
 # Hints
 
+## General
+
 Return values for the functions are of a specific type, so pay attention.
 
 ## 1. Define a named vector


depial:
A quick note: I used configlet to create the exercise, and it automatically filled in `examplar.R` instead of `exemplar.R`, so we should probably track down the source of that typo.

depial:
> Looks like config.json, line 32

Let's merge this first, then I'll correct the `config.json` entries.

depial:
Works on the site, but it took a long time to pass again. I'm hoping I just have to restart my computer or something.

colinleach:
I'm also seeing this on the Julia and Python tracks (Lasagna, in both cases, so nothing challenging)

colinleach:
I've posted an FYI to the Maintainers channel on Discord. It _may_ be related to Jeremy getting the Jiki stuff together - we also had glitches a year ago when the bootcamp launched.

PR #447, opened 2026-02-15
Change `examplar.R` to `exemplar.R`

depial:
As the title says. The changes took place in `config.json`, and `lasagna/meta/config.json` as well as `lasagna/meta/exemplar.R` 

I think I've changed all the necessary instances.


colinleach (approved):

colinleach:
A quick grep shows that there is also a [reference document](https://github.com/exercism/r/blob/main/reference/implementing-a-concept-exercise.md) that advises contributors to do this wrongly. A couple of occurrences there. Do you want to fix this while you have the PR open?

PR #451, opened 2026-02-17
Errors Concept

colinleach:
If this seems weird and a bit confusing, I'll blame R.

I'm assuming we'll use Factory Sensors as the exercise. Again, there's an old R draft that was ported to Julia.


colinleach:
I've dropped this to draft, because I just realized the sample code still needs formatting. Tomorrow.

depial (approved):
All looks good. I guess the only question that I found coming to mind is why `tryCatch` is beyond the scope of the `about.md`. That's probably because it seems more familiar to me due to my Julia/Python background, so it's not a major concern.

PR #454, opened 2026-02-22
bin/verify-exercises-in-docker

keiravillekode:
Adapted from generic-track

$ bin/verify-exercises-in-docker -i my-local-image hello-world

will use the local docker image `my-local-image`


keiravillekode:
I use this with proposed test runner changes, to verify all the example solutions still pass.

colinleach:
Thank you!

PR #456, opened 2026-02-25
Update ci.yml

depial:
Naive first shot at addressing the unnecessary CI checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,11 +6,19 @@
 name: Exercise tests
 
 on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - "exercises/**"
+      - "bin/run_tests.R"
+      - ".github/workflows/ci.yml"
+  pull_request:
+    paths:
+      - "exercises/**"
+      - "bin/run_tests.R"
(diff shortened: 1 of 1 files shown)


colinleach (approved):

colinleach:
Another 6+ minute build for s2, which is apparently "spherical geometry operators". I've tried to find out why it's being imported, with no success so far. At least it doesn't seem to be a dependency of either `lubridate` or `forcats`, so I don't think the previous PR is to blame (unless it was the R version update?).

depial:
I did notice that this time it was around 8 minutes before it even got the installation of packages.

There is this:

> Cache not found for input keys: Ubuntu 24.04.3 LTS-R version 4.5.2 (2025-10-31)-1-66ab042957c4cd0b8bf61ab0fdab80148629878ff657ec9558a4dfa39a87df3b, Ubuntu 24.04.3 LTS-R version 4.5.2 (2025-10-31)-1-

Which seems like it could be relevant. I'm just out of my depth when it starts to get to this kind of thing though.

colinleach:
I'm just reading [this](https://github.com/r-lib/actions/tree/v2/setup-r-dependencies). I think at the very least we need to upgrade `cache-version` to 2. The examples seem to use `r-lib/actions/setup-r@v2` rather than the big UUID, but I also don't know what I'm doing here.

More reading required...

PR #462, opened 2026-03-01
minor cleanup to Switch concept

colinleach:
The Switch concept is not appearing on the Exercism website: not visible on the syllabus tree, and entering the URL manually gets a 404.

These are mainly cosmetic changes, slightly useful in themselves but mainly intended to trigger the GH-Exercism sync.

### DIFF
--- a/concepts/switch/about.md
+++ b/concepts/switch/about.md
@@ -17,7 +17,7 @@ star <- function(type) {
 [1] "red dwarf"
 ```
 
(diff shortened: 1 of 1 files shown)


colinleach:
I'm hoping to PR Factory Sensors (for the Errors concept) today or tomorrow. It might be easiest to do something trivial to Blackjack within that, instead of an entire separate PR.

PR #463, opened 2026-03-02
Update broken link in docs

depial:
Just a slight rename of a link to match the definition. Also, this may help push the website to see the `switch` concept.

### DIFF
--- a/concepts/switch/about.md
+++ b/concepts/switch/about.md
@@ -29,7 +29,7 @@ The `switch` in base R is quite limited: it will only do an exact match to a sin
 This seemed reasonable when R was first released in 1993, but needs improvement for modern usage.
 
 ~~~~exercism/note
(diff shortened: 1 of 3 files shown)


colinleach (approved):

colinleach:
I found another problem. The [docs](https://dplyr.tidyverse.org/reference/case_match.html) say  `case_match` was recently deprecated, so I'll need to rewrite that bit of the concept.

> case_match() is deprecated. Please use [recode_values()](https://dplyr.tidyverse.org/reference/recode-and-replace-values.html) and [replace_values()](https://dplyr.tidyverse.org/reference/recode-and-replace-values.html) instead, which are more powerful, have more intuitive names, and have better safety. In addition to the familiar two-sided formula interface, these functions also have from and to arguments which allow you to incorporate a lookup table into the recoding process.

PR #466, opened 2026-03-08
replace case_match with recode_values

colinleach:
History:
- The `switch` concept was written February 1st
- Version 1.2.0 of `dplyr` was released February 4th
- The `case_match()` function which I had so lovingly documented is now deprecated
- The `recode_values()` function is now recommended instead
- I feel a bit aggrieved about this

This PR updates the `switch` concept and `blackjack` exercise to reflect the library changes. This is important, because the test runner already uses dplyr 1.2. I'm not sure about CI, but we'll find out when I create this PR.


colinleach:
**N.B.** Reviewing this is _not_ something for this evening. It can wait for future days!

PR #469, opened 2026-03-11
easy syncs

colinleach:
Despite the branch name, this doesn't add any templates. Just some preliminary housekeeping.

I'd intended to include `allergies`, but that's so out of date that I'm struggling to find any relationship between the current tests and the `problem-specifications`. I'll sort it out later.

Anyone doing a first j2 template might be well advised to pick an exercise with no nesting in the JSON. The templates can certainly handle nesting, it just needs more thought.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,9 @@ bin/configlet.exe
 .RData
 .Rhistory
 *.Rproj
+
+bin/__pycache__/
+.idea/
+.venv/
+
+.problem-specifications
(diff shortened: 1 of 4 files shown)


colinleach:
And when we're bored with templates, we can go back to the [syllabus](https://exercism.org/tracks/r/concepts) which Bethany should now be able to see. Still plenty to do, but it's making progress (sporadically).

colinleach:
Progress tends to go in bursts. I don't think either of is the slow-and-steady type of personality.

Issue #465, opened 2026-03-08
Sync of practice exercises


colinleach:
I just ran `configlet sync` and found the results a bit alarming.

<details>
<summary>Sync Results</summary>

```
[warn] docs: instructions unsynced: killer-sudoku-helper
[warn] allergies: missing 1 test case
       - list when: -> no allergen score parts without highest valid score (93c2df3e-4f55-4fed-8116-7513092819cd)
[warn] anagram: missing 3 test cases
       - does not detect an anagram if the original word is repeated (630abb71-a94e-4715-8395-179ec1df9f91)
       - handles case of greek letters (a6854f66-eec1-4afd-a137-62ef2870c051)
       - different characters may have the same bytes (fd3509e5-e3ba-409d-ac3d-a9ac84d13296)
[warn] bob: missing 1 test case
       - multiple line question (2c7278ac-f955-4eb4-bf8f-e33eb4116a15)
[warn] collatz-conjecture: missing 2 test cases
       - zero is an error (2187673d-77d6-4543-975e-66df6c50e2da)
       - negative value is an error (ec11f479-56bc-47fd-a434-bcd7a31a7a2e)
[warn] crypto-square: missing 2 test cases
       - normalization results in empty plaintext (aad04a25-b8bb-4304-888b-581bea8e0040)
       - 54 character plaintext results in 8 chunks, the last two with trailing spaces (33fd914e-fa44-445b-8f38-ff8fbc9fe6e6)
[warn] flatten-array: missing 4 test cases
       - consecutive null values at the front of the array are omitted from the final result (bc72da10-5f55-4ada-baf3-50e4da02ec8e)
       - consecutive null values in the middle of the array are omitted from the final result (6991836d-0d9b-4703-80a0-3f1f23eb5981)
       - 6 level nested array with null values (dc90a09c-5376-449c-a7b3-c2d20d540069)
       - all values in nested array are null (51f5d9af-8f7f-4fb5-a156-69e8282cb275)
[warn] hamming: missing 6 test cases
       - disallow first strand longer (b9228bb1-465f-4141-b40f-1f99812de5a8)
       - disallow second strand longer (dab38838-26bb-4fff-acbe-3b0a9bfeba2d)
       - disallow left empty strand (db92e77e-7c72-499d-8fe6-9354d2bfd504)
       - disallow empty first strand (b764d47c-83ff-4de2-ab10-6cfe4b15c0f3)
[...]

PR #473, opened 2026-03-11
[Add template] Hamming

colinleach:
The test file seems to have got shortened, but it's not obvious why. Everything is in a different sequence, and none of the old tests look track-specific.

After an initial panic, I'm now pretty relaxed about this. If the canonical tests are good enough for 60+ other tracks, I can live with them. Unless there's a specific reason like vectorization, which isn't the case here.

### DIFF
--- a/exercises/practice/hamming/.meta/tests.toml
+++ b/exercises/practice/hamming/.meta/tests.toml
@@ -1,6 +1,13 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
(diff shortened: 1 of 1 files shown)


BethanyG:
I am ... just going to ask.  I see that there were [6 missing test cases for `Hamming`](https://github.com/exercism/r/issues/465) when the sync was run.  Did you make sure to pull in all the changes from main?  If so -- I am with @depial -- we can trust the canonical data, and not worry about the redundancy.  But let's make sure that all the tests needed are indeed in the `tests.toml`.

I mean, we can always adjust and re-generate (yay!), so no real harm if we discover missing stuff later.

PR #476, opened 2026-03-12
Add roman-numerals and reverse-string templates

depial:
This updates `roman-numerals` and `reverse-string` by syncing with `canonical-data` and adding test `template.j2`s.

Note: I did not include two new tests for `reverse-string` which require grapheme cluster manipulation. These require either `stringr` or `stringi`, and I'm not sure if we want to enforce the use of a library or if we want to keep everything passable in Base R. I think this would be a good time to resolve this question.


colinleach:
There are only 82 completions, which may be relevant to know.

Also, this exercise has pretty thorough Approaches docs (_I can't imagine who wrote them..._). Use of `stringi` is recommended there, but they now need updating for `stringr`.

PR #477, opened 2026-03-13
[Practice Exercise] Add Bottle Song

colinleach:
Notionally replaces `beer-song`, though in practice they have little in common.

There's a couple of small, unintended deletions in `config.json` that slipped through, but I don't think they will matter (I hope!)

### DIFF
--- a/exercises/practice/bottle-song/.meta/config.json
+++ b/exercises/practice/bottle-song/.meta/config.json
@@ -1,0 +1,19 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "files": {
+    "solution": [
+      "bottle-song.R"
+    ],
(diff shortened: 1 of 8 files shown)


exercises/practice/bottle-song/.meta/config.json
```diff
@@ -0,0 +1,19 @@
...
+      "bottle-song.R"
+    ],
+    "test": [
+      "test_bottle-song.R"
+    ],
+    "example": [
+      ".meta/example.R"
+    ]
+  },
+  "blurb": "Produce the lyrics to the popular children's repetitive song: Ten Green Bottles.",
```
depial: Is this a popular children's song? I must have missed it ;D

PR #481, opened 2026-03-13
[Add template] Isogram

colinleach:
Another simple one.

### DIFF
--- a/exercises/practice/isogram/.meta/template.j2
+++ b/exercises/practice/isogram/.meta/template.j2
@@ -1,0 +1,11 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.canonical_ref() }}
+
+{{ macros.header() }}
+
+{% for case in cases -%}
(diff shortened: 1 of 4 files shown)


BethanyG:
Did you update the test cases from `problem-specs` for this already ... or are we going to do those in a different step?

BethanyG:
There are a LOT of unsynced changes for the track -- so it is understandable if the process is a tad chaotic.  I just don't want them to go by the wayside, so probably going to be annoying and ask to remind folx.  I just didn't see a sync PR fly past, and didn't see the updates here.  So.  😄 

Might also be a good idea to maybe re-run configlet sync when we get this first batch done and sorted.  Just to make sure we did indeed update things.

colinleach:
I may give templates a break for a day or two. I've been aiming to get `factory-sensors` sorted out for the Errors concept for about the last few months, and I also need to get Randomness/Captain's Log finished. The latter is _really_ important in a language aimed at statistics and data modelling!

PR #493, opened 2026-03-18
[Add concept] Functions

colinleach:
I'll park this draft here for now, but it's too early for a detailed review.

I'm currently working on `functional-programming`, with pipes, apply functions, `purrr`, recursion, etc. It's probably best to wait until we can review them as a pair, and see what fits where.


colinleach:
See #501 for a discussion of concept exercises, for this and `functional-programming`.

depial:
The `about.md` looks good after a cursory look over it.

I also agree that `locomotive-engineer` might not be a good fit for an exercise. I can try to think of some type of adaptation or something else entirely over the weekend, though.

Let me know if you have any other ideas for exercises for this concept.

colinleach:
I may fall back on [clojure/coordinate-transformation`](https://github.com/exercism/clojure/tree/main/exercises/concept/coordinate-transformation) if we can't think of anything better. It's focused on closures.

colinleach:
The exercise Intro was copied to this About, then a shortened version added to this Intro. If that makes sense?

I'll update the exercise with the shortened Introduction. Though the docs are relatively terse by my expansive standards!

concepts/functions/about.md
```diff
@@ -0,0 +1,243 @@
...
+This section will make more sense once we reach the [Functional Programming][concept-funcprog] Concept.
+Below is a preview, using [`sapply()`][ref-sapply] to square each number in a range:
+
+```R
+sapply(1:5, \(x) x ^ 2)
+#> [1]  1  4  9 16 25
+```
+
+That is not a very useful example, because `(1:5) ^ 2` returns the same result more simply, but it illustrates how we define a function without bothering to think of a name for it.
+
```
depial: We could/should also make a call back to the anonymous function that was used in the example in the previous section. I suppose we could even drop this example in favor of that call back, since this example needs a lot of foreshadowing / hand waving.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,217 @@
...
+
+This section will make more sense once we reach the [Functional Programming][concept-funcprog] Concept.
+Below is a preview, using `sapply()` to square each number in a range:
+
+```R
+sapply(1:5, \(x) x ^ 2)
+#> [1]  1  4  9 16 25
+```
+
+That is not a very useful example, because `(1:5) ^ 2` returns the same result more simply, but it illustrates how we define a function without bothering to think of a name for it.
```
depial: Same comment as in the `about.md`

concepts/functions/links.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+      "url": "http://adv-r.had.co.nz/Functions.html",
+      "description": "Functions chapter in Advanced R."
+    }
+  ]
+
```
depial: GH newline complain

depial (approved):
This looks good to me! This may be shorter than other concepts, but I can't imagine there's much more to squeeze in here.

concepts/functions/links.json
colinleach: There was a newline, but a different formatting issue confused it.

concepts/functions/about.md
colinleach: I dropped the rather tortured example and referred back. I also fixed some mangled indents in the previous example.

PR #495, opened 2026-03-20
[Test Templates]: Diamond, ETL, Grains, and Resistor-Color

BethanyG:
Added JinJa templates, regenerated test files, and re-tested example files:

- [x] Diamond
- [x] ETL
- [x] Grains
- [x] Resistor-Color

Wow, did `Grains` and `Resistor-Color` try my patience.  Nesting with identical keys in problem specs.  Yuck.
OTOH, `ETL` was .. fun?  I learned all about `if not loop.last`.  😉

### DIFF
--- a/exercises/practice/diamond/.meta/template.j2
+++ b/exercises/practice/diamond/.meta/template.j2
@@ -1,0 +1,16 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.canonical_ref() }}
+
+{{ macros.header() }}
+
+{% for case in cases -%}
+test_that("{{ case["description"] }}", {
+  input <- "{{ case["input"]["letter"] }}"
+  expected <- c{{ case["expected"] | list_to_tuple }}
+  expect_equal(
+    diamond(input),
+    paste(collapse= "\n", expected)
+  )
+})
+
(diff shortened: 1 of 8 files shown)


colinleach (approved):

depial:
> OTOH, `ETL` was .. fun? I learned all about `if not loop.last`.

Nice! I'm wondering if we should keep some kind of catalog of templates that provide an example of useful structures?

colinleach:
> keep some kind of catalog of templates

A list of all the templates that caused us pain, to reduce the risk of repeating that pain. I can see the merits of this!

BethanyG:
> _Nice! I'm wondering if we should keep some kind of catalog of templates that provide an example of useful structures?_

**A yes to this.** (_I'll try to put something formal together-ish at some point_)  😄  

That's (sort of) what I've been doing when I make templates:  assess if I can copy a previous successful template (or structure in a template) I've made and just make spot changes.  Worked (thank goodness!!) for `resistor-color`, since it was nested in the same way `grains` was.

Also?  Patterns that are useful and also repeat can be made into macros.  Of course by the time I get there, we'll have made all the exercise templates.  But still.

PR #496, opened 2026-03-26
[Practice exercise] Baffling Birthdays

colinleach:
A bit off the priorities, but an exercise that covers random dates and probability estimates seems too good to miss.

Not ideal for templating, as so many of the results being tested are non-deterministic. Hence, the `canonical-data` isn't very canonical! I did what I could...

One test I might think of adding is an explicit check that the random birthdates are actually dates (not strings). That's implicit in some other tests, but I'm not sure how useful the error messages will be.

Difficulty 6 is the same as Julia, C# and Elixir.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1064,6 +1064,14 @@
         "difficulty": 6
       },
       {
+        "slug": "baffling-birthdays",
+        "name": "Baffling Birthdays",
+        "uuid": "fd94d735-3d0e-4c6d-8b65-5fe170e00f1e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 6
+      },
+      {
         "slug": "zebra-puzzle",
         "name": "Zebra Puzzle",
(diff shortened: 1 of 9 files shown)


colinleach:
@BethanyG, does this exercise seem like it would be a useful addition to the Python track?

colinleach:
The power of caching: the first CI run failed in 35 minutes, the second passed in 4 minutes.

The [CI workflow](https://github.com/exercism/r/blob/main/.github/workflows/ci.yml) lists packages separately, in an attempt to limit how many to build. I'd be interested in an experiment replacing most of those with `tidyverse`, to see how that compares.

depial (approved):
lg2m!

depial:
> I'd be interested in an experiment replacing most of those with tidyverse, to see how that compares.

This seems reasonable to me (with a nod to my limited knowledge in the area). I was wondering if the cache had cleared and then there was some issue loading tidyverse, but I guess it's not even loaded.

PR #497, opened 2026-04-19
[New Practice Exercise] Flower Field

colinleach:
I'll get back to syllabus stuff later, but while I have limited time/energy I'll try to add a few practice exercises.

I've set this to difficulty 4. Most languages (including Python) are 4 or 5. Quite why Julia is a 2 is a question for another day.

Plans to stick to base R fell by the wayside. Tidyverse functions are _so_ much easier.

I learned more about R while writing the example, which should be useful when I eventually create the concepts. Manipulating matrices created from outer products of complex-valued vectors wasn't originally in my plans!

### DIFF
--- a/config.json
+++ b/config.json
@@ -1078,6 +1078,14 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 8
+      },
+      {
+        "slug": "flower-field",
+        "name": "Flower Field",
+        "uuid": "031f47c6-e7bc-4b69-aace-2f0b2f6620d1",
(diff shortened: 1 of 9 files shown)


depial (approved):
Sounds complex! :D

I'm also not exactly sure why Juila is a 2, but maybe it just carried over from when we switched from Minesweeper? That said, I don't remember finding it too difficult... But I guess I made use of comprehensions and dictionary functions, so I would not be against bumping it up in difficulty.

PR #503, opened 2026-05-10
[New Concept] Matrices and Arrays

colinleach:
Quite a fitful path to this PR: I started it weeks ago, stopped when it no longer seemed important enough, then restarted when I wanted to refer to it in the Dataframes concept. You'll notice my residual hesitancy in the text!

We have a couple of options for the exercise:
1. Use a cut-down port of `julia/exercism-matrix`.
2. Leave it as a documentation-only concept that we can link to.

In the first case we'd want a much-abbreviated `introduction.md`, in the second case we might as well make it the same as `about.md` (which would maybe be hard to reach otherwise???)

### DIFF
--- a/concepts/matrices-arrays/about.md
+++ b/concepts/matrices-arrays/about.md
@@ -1,0 +1,370 @@
+# About
+
+When handling collections of homogeneous atomic data (e.g. all numeric or all character), the following structures are available in R:
+
+- `vector` for 1-dimensional data, accessed with a single index.
+- `matrix` for 2-dimensional, rectangular (not ragged) data.
+- `array` for the general N-dimensional case. A `matrix` is just an `array` with N=2.
+
+~~~~exercism/note
+**For Python programmers**: This section may look familiar. 
+
+Any resemblance to NumPy is not coincidental (but R is slightly older, and has all this in the base language).
+~~~~
+
+## Matrices
+
+A matrix is essentially just a vector, plus row and column counts to define the shape.
+
+There is a [`matrix()`][ref-matrix] function to create them:
+
+```R
+# data already available 
+matrix(1:6, nrow=2, ncol=3)
+#>      [,1] [,2] [,3]
+#> [1,]    1    3    5
+#> [2,]    2    4    6
+
+# empty matrix to be filled in later
+matrix(nrow=2, ncol=3)
+#>      [,1] [,2] [,3]
+#> [1,]   NA   NA   NA
+#> [2,]   NA   NA   NA
+```
+
+Note that the data is stored in column-major order (i.e. down then across).
+
+Find the dimensions of a matrix with [`dim()`][ref-dim] for both rows and columns, [`nrow()`][ref-nrow] or [`ncol()`][ref-nrow] to get a single dimension.
+Using [`length()`][ref-length] will just return the length of the underlying vector (6 in the example above).
+
+Some other functions create matrix output, for example [`str_split_fixed()`][ref-str_split_fixed].
+
+```R
+library(stringr)
+
+str_split_fixed(c("split some words", "and some more"), " ", 3)
+#>      [,1]    [,2]   [,3]   
+#> [1,] "split" "some" "words"
+#> [2,] "and"   "some" "more" 
+```
+
+## Matrix indexing
+
+This works in all the same ways as vector indexing, except for the need to specify row(s) and column(s), in that order.
+Leave a dimension empty to get the entire row/column.
+
+Using the same matrix as above:
+
+```R
+m <- matrix(1:6, nrow=2, ncol=3)
+m
+#>      [,1] [,2] [,3]
+#> [1,]    1    3    5
+#> [2,]    2    4    6
+
+m[2, 3]    # row 2, col 3
+#> [1] 6
+
+m[1:2, 3]  # rows 1:2, col 3
+#> [1] 5 6
+
+ m[2, ]    # all of row 2
+#> [1] 2 4 6
+```
+
+Note that a subset with a single row and/or column will be returned as a vector, not a matrix.
+
+```R
+class(m)
+#> [1] "matrix" "array" 
+class(m[2,])
+#> [1] "integer"
+```
+
+This type of `dimension reduction` is not always wanted, especially in a function pipeline that expects matrix inputs.
+
+Specifying `drop=FALSE` will guarantee a matrix return value from matrix input.
+Note that a matrix with only one row or one column is _not_ the same type as a vector.
+
+```R
+m[2, , drop = FALSE]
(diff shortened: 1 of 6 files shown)


colinleach:
My next priority is to finish the Dataframes concept. I see that as the absolute must-have for launching an R syllabus, as it's central to most of the Tidyverse, and most of what R is used for these days (used a lot: R rose to 8th place in the most recent Tiobe index).

@BethanyG, do you have any favorite exercise ideas for Python + NumPy + Pandas? R has the equivalent of all those in the test runner, so this is our first chance to dip a toe in the Data Science water.

colinleach:
I need to refer to this from the Dataframes concept, and avoiding a broken link now seems more important than adding an exercise here.

For now, I've made the Intro the same as the About (LinAlg bits and all!). I posted a [question](https://forum.exercism.org/t/confusion-about-concept-docs-again/53692) on the forum about this, no answers yet.

colinleach:
It's remarkable looking at the history and finding that R was once seen as a full replacement for Matlab, complete with automated script translation and lots of "converting from Matlab" documentation. Even ignoring Octave, R was squeezed between NumPy and Julia in the OSS space, while serious Matlab users _really need_ Simulink and are willing to pay lots of $$$ for it.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+#> [2,]    2    4    6
+
+# empty matrix to be filled in later
+matrix(nrow=2, ncol=3)
+#>      [,1] [,2] [,3]
+#> [1,]   NA   NA   NA
+#> [2,]   NA   NA   NA
+```
+
+Note that the data is stored in column-major order (i.e. down then across).
```
depial: Maybe a bit more information than the current parenthetical comment could be useful here? I could see either a [link](https://en.wikipedia.org/wiki/Row-_and_column-major_order), or a mention to other languages (e.g. `like Julia... and as opposed to Python...`), being added.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+
+Find the dimensions of a matrix with [`dim()`][ref-dim] for both rows and columns, [`nrow()`][ref-nrow] or [`ncol()`][ref-nrow] to get a single dimension.
+Using [`length()`][ref-length] will just return the length of the underlying vector (6 in the example above).
```
depial: I found the `the length of the underlying vector` an interesting way of looking at this. I would have naturally gravitated towards something like `the total number of elements`. The former seems more "constructive", while the latter seems more "descriptive" (for lack of better terms).

Edit: I suppose the current way is more descriptive than "total elements" in a way since the underlying memory structure is indeed sequential like a vector. I do quite prefer how the sentence currently stated with `underlying vector`.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+
+This type of `dimension reduction` is not always wanted, especially in a function pipeline that expects matrix inputs.
+
+Specifying `drop=FALSE` will guarantee a matrix return value from matrix input.
+Note that a matrix with only one row or one column is _not_ the same type as a vector.
+
+```R
+m[2, , drop = FALSE]
+#>      [,1] [,2] [,3]
+#> [1,]    2    4    6
```
depial, line 93: It could be nice to add `class(m[2, , drop = FALSE])` and output here to mirror the text, but I don't think it's necessary either since the output from `m[2, , drop = FALSE]` is likely enough.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+
+```R
+m[, 2] <- c(10, 20)
+m
+#>      [,1] [,2] [,3]
+#> [1,]    1   10    5
+#> [2,]    2   20    6
+```
+
+## Adding and deleting rows/columns
```
depial, line 116: There doesn't seem to be any mention of *deleting* row/columns in this section.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+- [Linear Algebra Basics][concept-linalg-basics] in Julia
+- [Linear Equation Solving][concept-lin-eq-solving] in Julia
+
+[concept-linalg-basics]: https://exercism.org/tracks/julia/concepts/linear-algebra-basics
+[concept-lin-eq-solving]: https://exercism.org/tracks/julia/concepts/linear-equation-solving
+~~~~
+
+### Diagonals
+
+The main diagonal of a _square_ matrix runs from top left (`A[1,1]`) to bottom right.
```
depial: For extra clarity/verbosity, this could be slightly rephrased as: 

> The main diagonal of a _square_ (`N x N`) matrix runs from top left (`A[1,1]`) to bottom right (`A[N,N]`).

I'll also note that, before, lowercase `m` was used for the matrix, while, from here on, it's uppercase `A`.

concepts/matrices-arrays/about.md
```diff
@@ -0,0 +1,375 @@
...
+# same calculation, different syntax
+m ^ 2
+#>      [,1] [,2] [,3]
+#> [1,]    1    9   25
+#> [2,]    4   16   36
+```
+
+Explaining the various other types of vector and matrix multiplication is beyond our scope.
+
+In very brief summary:
```
depial: I feel like these two sentences naturally combine into one:

> Explaining the various other types of vector and matrix multiplication is beyond our scope, however, in very brief summary:

depial (approved):
Looks good! I liked learning about the "Go figure" aspects of the language :D

Just a note: I saw that you'd stated the `Introduction` was just a copy of the `About`, so I haven't read through it.

Approving since most things are just optional comments/possible suggestions (with the exception of the pointer to the section about `Adding and deleting rows/columns`, which I think should be addressed).

concepts/matrices-arrays/about.md
colinleach: I tend to steer away from discussing `class` most of the time, for reasons that will become clearer when I finish the Class concept.

concepts/matrices-arrays/about.md
colinleach: A very valid point! I added a short bit (use negative indexing, like vectors).

concepts/matrices-arrays/about.md
colinleach: I've expanded it a bit. I thought about explaining LinAlg conventions on case, but decided brevity is better. I don't want to seem too encouraging here!

colinleach:
Now you've read these docs, I guess you understand why I backed away from pairing this with a complex and difficult exercise. Mention it, move on...

PR #507, opened 2026-05-29
[New Concept] Regular Expressions

colinleach:
Another incomplete draft. It can be finished once we have `regular-chatbot` to pair with it. I also need to read it tomorrow with fresh eyes: I'm confident there are typos!

I went though a period of thinking this concept is non-urgent, then I found I kept referring to it in other concepts. Without it, we'll have lots of broken links scattered around the syllabus.

### DIFF
--- a/concepts/regular-expressions/.meta/config.json
+++ b/concepts/regular-expressions/.meta/config.json
@@ -1,0 +1,7 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [],
+  "blurb": "Regular expressions are available in Base R, but much easier to work with using the string library."
+}
--- a/concepts/regular-expressions/about.md
+++ b/concepts/regular-expressions/about.md
@@ -199,6 +199,8 @@ Matrices do not fit well with the Tidyverse ecosystem, despite being part of Bas
 In quick summary: treat matrices like vectors with 2 indices, in the order `[rows, cols]`.
 Leave either index blank to get the whole row/column.
 
+When getting a single row `r` with `m[r,]` or a single column `c` with `m[,c]`, the value returned is flattened to a vector by default.
+
 ```R
 > amounts <- recipe |> str_match("(?<wt>\\d+g) .* (?<vol>\\d+ml)")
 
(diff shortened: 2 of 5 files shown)


colinleach:
I've copied the `introduction.md` from `regular-chatbot`, and added updates to `about.md`. I think (hope?) it's about ready for review.

depial (approved):
I see nothing wrong here! However, I'm coming from a learner's perspective since I'm not really well versed in the `stringr` library.

The one question that came to mind is that I think I remember that there is a `perl` option in Base R which is needed for the PCRE2 engine to be used. Does `stringr` use that automatically? If so/not, maybe this info could be useful (or maybe it could be superfluous).

colinleach:
Yes, I had already wondered what (if anything) I should say about the Base R engines. TRE is the default, PCRE2 can be selected with `perl = TRUE`. The `stringi` library standardized on ICU (only), and `stringr` is built on top of that. 

This isn't my expertise, but as far as I can tell ICU is close to PCRE2, _maybe_ slightly newer and with a formal standards body behind it. No idea about TRE, it seems to be have been an ever-changing mess from the early years of regex programming.

On reflection, my inclination (that could be argued against!) is to leave the Base R stuff at "_Here be Dragons - avoid!_"

My other slight concern is that I skipped over handling non-English locales. TBH, I'm not sure I'm competent to write that.

depial:
> On reflection, my inclination (that could be argued against!) is to leave the Base R stuff at "Here be Dragons - avoid!"

I quite like Base R, due to that being the only option here for a long time, but it's also probably because I haven't really gotten to know `tidyverse` that well, so I assume it's more of a case of just not being aware of what I'm missing :D

> My other slight concern is that I skipped over handling non-English locales.

I'd assume this could be included as an aside about Unicode properties without much trouble. As there is the drive to translate Exercism into many languages, maybe the language specific details could be passed on to those translators?

colinleach:
Locales are _complicated_. I already knew that in a regex context, but I just spent a few hours fighting it in a date-time context (trying to implement `booking-up-for-beauty`). Fair play to R, they recognize the complexity and try to work with it, but `expect_equal()` is going to struggle if the student code returns a string in their system language.

PR #509, opened 2026-05-30
[Concept Exercise] booking-up-for-beauty

colinleach:
I'm not happy with this yet (a very direct port from Julia):

- I'm worried about task 4, with locale-aware string formatting.
- I'd like to do more on datetime deltas: maybe simplify something from [Swift Scheduling](https://github.com/exercism/r/tree/main/exercises/practice/swift-scheduling)? 

Hence no hints yet, they may change later.


depial:
I'll probably go through this in more detail tomorrow, but as a first thought on your two points, you could kill two birds with one stone: Replace Task 4 with a datetime delta task. For example, via a `reschedule` function (storyline of  customers needing to reschedule appointments).

That said, I don't feel too uncomfortable with the use of locales in Task 4, since I think you cover that point well in the introduction and it would be easy to give an explicit hint. So, I'm currently open to keeping or dropping as you see fit.

colinleach:
It sounds like we're having similar thoughts about the tasks. I'll aim to do something along these lines tomorrow.

colinleach:
I've commented out the old tasks 4 and 5, in multiple files: 4 because of locale issues as discussed, 5 because I don't have a way to test it without putting the entire solution in the tests file. We'll need to remember to delete these once we agree on the exercise structure - don't merge big blocks of outdated stuff.

The new task 4 practices day-of-week stuff, but is there mainly as a big hint for task 5.

The new task 5 is a step up in difficulty, but re-using tasks 1 and 4 will help a lot. I got caught out by `ifelse()` stripping datetimes back to a bare number without attributes (crap!), so I added a warning in the hints.

exercises/concept/booking-up-for-beauty/.docs/instructions.md
```diff
@@ -0,0 +1,80 @@
...
+
+```R
+# a Friday
+day_of_week(make_datetime(2026, 06, 05, 12, 30, 00))
+#> [1] 5
+```
+
+## 5. Reschedule an appointment
+
+The customer is no longer able to make the original appointment, and has asked to reschedule to the next Friday at the same time of day.
```
depial: I might change the story just a bit to make the situation more programmatic. Maybe something like:

> When a customer needs to reschedule an appointment, this is automatically made for the next Friday at the same time of day.

A parenthetical comment could be made to the effect `since this is the slowest day of the week` if more context is desired.

exercises/concept/booking-up-for-beauty/.docs/hints.md
```diff
@@ -0,0 +1,31 @@
...
+
+## 4. Check the day of the week for an appointment
+
+- Check the Introduction: Monday is not the _default_ day 1 in lubridate.
+
+## 5. Reschedule an appointment
+
+- The input is in the same format as Task 1.
+- What day is the original appointment for?
+  Task 4 can help with this.
```
depial: Since these are relative times, I'm not sure Task 4 needs to be explicitly mentioned. Personally, I found it easier/more natural to think (and operate) in terms of Sunday = Day 1, and, come to think of it, it's probably smarter to use Friday = Day 1:

```R
reschedule <- function(appointment) {
  app <- schedule_appointment(appointment)
  weekday <- wday(app, week_start=5)
  app + days(8 - weekday)
}
```

exercises/concept/booking-up-for-beauty/.docs/hints.md
```diff
@@ -0,0 +1,31 @@
+# Hints
+
+## 1. Parse all-number appointment date
+
+- Lubridate has many [helper functions][ref-parse] for datetime parsing.
+
+## 2. Check if an appointment has already passed
+
+- When [getting the current time][ref-now], remember to use `UTC` and not your local timezone.
```
depial, line 9: I'm trying to think of how to test that the student uses UTC, since the tests pass with `now()`... 

Edit: I don't think it's possible for this task to enforce UTC, since lubridate handles the time zones when doing a simple comparison.

depial (approved):
Looks good on the whole! Just a couple minor suggestions.

depial:
One thing I didn't do was to check if the `introduction.md` could/should be trimmed down... If you don't mind, I'll leave that up to you since you are more familiar with how everything fits together here and my brain is getting a bit too foggy for the day :sweat_smile: 

(However, I did try to look it over to check for corrections/differences/etc)

exercises/concept/booking-up-for-beauty/.docs/hints.md
depial: Just to clarify my point, when I first did the exercise, I didn't even think about connecting it to Task 4, and I did it using the standard Sunday as the start of the week. However, seeing your hint later led me to the Friday solution, so maybe we could:
- just reword the hint to be more general?
- add another hint pointing towards the Friday possibility? 
- I also wouldn't mind just leaving it as is, but my initial concern is that explicitly bringing Task 4 into the mix could confuse some people who would have otherwise used the Sunday route.

exercises/concept/booking-up-for-beauty/.docs/hints.md
colinleach: I think you can tell I get confused by locale issues!

I don't think this is a big deal, so trying to test `tz` is more trouble than it's worth.

colinleach:
> One thing I didn't do was to check if the introduction.md could/should be trimmed down.

I guess it probably should be, but:
- I can't immediately see how.
- This comes quite late in the syllabus, so less interested students probably won't get this far.

I'm getting tired (and lazy) myself, so I'm inclined to merge as-is, and assume we'll probably make a few changes once more of the syllabus is in place on the website. Plus even more changes once we get complaints from students!

depial:
Looks good on the web editor! I'm confused why we're allowed to merge concept exercises here without a concept, but I'm not going to complain :D

PR #510, opened 2026-06-01
[Concept Exercise] Captains Log

colinleach:
Replaces #273.

I'd forgotten I wrote this back in 2023, but that needed a pretty major rewrite. Tests are now based on the Julia version, and the Intro is from #490.

I suspect we'll need to tweak the testing once it's running on the website. I reduced the number of repeats relative to Julia, to try and minimize timeouts. I don't currently know how slow it will be, nor what the false-negative rate will be.

### DIFF
--- a/exercises/concept/captains-log/.docs/hints.md
+++ b/exercises/concept/captains-log/.docs/hints.md
@@ -1,0 +1,24 @@
+# Hints
+
+## 1. Generate a random planet
+
+- This looks like [sampling][ref-sample] from an input vector.
+
+## 2. Generate a random starship registry number
+
+- The numeric part is sampling from a range.
+- There are several ways to construct the string, including [`sprintf()`][ref-sprintf] and [`str_glue()`][ref-str_glue].
+
+## 3. Generate a random stardate
+
+- We want a floating-point value with [uniform distribution][ref-runif] within lower and upper limits.
+
+## 4. Pick some random starships from a list
+
+- More sampling from an input vector.
+- Another approach is to shuffle the starships, then return the number needed from the start of the resulting vector.
+
+[ref-sample]: https://www.rdocumentation.org/packages/base/versions/3.3.0/topics/sample
+[ref-sprintf]: https://www.rdocumentation.org/packages/base/versions/3.3.0/topics/sprintf
+[ref-runif]: https://www.rdocumentation.org/packages/stats/versions/3.5.2/topics/Uniform
+[ref-str_glue]: https://stringr.tidyverse.org/reference/str_glue.html
--- a/exercises/concept/captains-log/.docs/hints.md
+++ b/exercises/concept/captains-log/.docs/hints.md
@@ -3,6 +3,7 @@
 ## 1. Generate a random planet
 
 - This looks like [sampling][ref-sample] from an input vector.
+- Duplicates are possible, so sample with replacement.
 
 ## 2. Randomize a fleet of starships
 
--- a/exercises/concept/captains-log/.docs/hints.md
+++ b/exercises/concept/captains-log/.docs/hints.md
@@ -5,14 +5,14 @@
 - This looks like [sampling][ref-sample] from an input vector.
 - Duplicates are possible, so sample with replacement.
 
-## 2. Randomize a fleet of starships
-
-- The Introduction discusses shuffling.
-
-## 3. Generate a random starship registry number
+## 2. Generate a random starship registry number
 
 - The numeric part is sampling from a range.
 - There are several ways to construct the string, including [`sprintf()`][ref-sprintf] and [`str_glue()`][ref-str_glue].
+
+## 3. Randomize a fleet of starships
+
+- The Introduction discusses shuffling.
 
 ## 4. Generate a random stardate
 
--- a/exercises/concept/captains-log/.docs/hints.md
+++ b/exercises/concept/captains-log/.docs/hints.md
@@ -4,19 +4,19 @@
 
 - This looks like [sampling][ref-sample] from an input vector.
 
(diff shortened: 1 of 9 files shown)


exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -0,0 +1,56 @@
...
+#> "NCC-1947"
+```
+
+## 3. Generate a random stardate
+
+What's the use of a log if it doesn't include dates?
+
+A stardate is a floating point number. The adventures of the _Starship Enterprise_ from the first season of _The Next Generation_ take place between the stardates 41000.0 and 42000.0. The "4" stands for the 24th century, the "1" for the first season.
+
+Implement the function `random_stardate` that returns a floating point number between 41000.0 (inclusive) and 42000.0 (exclusive).
```
depial: I saw that the `runif` gives inclusive values, and so does a range/sequence, so we could either:
1. Mimic task 2: `from 41000.0 to 42000.0 (inclusive)`
2. Drop both any mention of inclusive/exclusive in the instructions

If I'm remembering correctly, this addition in Julia came from someone arguing from a pure math standpoint, but it seems irrelevant here.

exercises/concept/captains-log/captains-log.R
```diff
@@ -0,0 +1,15 @@
+random_planet <- function() {
```
depial: This function should be named `random_planet_class`.

depial (approved):
Looks good in general. However, I'm wondering if we could rephrase Task 1 and/or Task 4 to get more out of the `introduction.md`. 

I would like to push the use of `replace=TRUE` at least and potentially use the shuffle feature of `sample` if possible.

As an example, we could add an argument for the number of planets in Task 1 and then test for numbers greater than the length of the list. For Task 4, we could just ask for the list of ships to be returned shuffled (and get rid of the argument for the number of ships). Solutions could look like:

```R
random_planet_class <- function(number_needed) {
    sample(c('D', 'H', 'J', 'K', 'L', 'M', 'N', 'R', 'T', 'Y'), number_needed, replace=T)
}

pick_starships <- function(starships) {
  sample(starships)
}
```

If we move in that direction, I could see a reshuffle of the Task numbers, with Task 1 and Task 4 swapping locations. (I would suppose `pick_starships` could potentially use a different name too...)

Whichever way we go, the `introduction.md` could maybe use some trimming down: `Normal/Gaussian distribution` at a minimum and the section on `shuffling` (if we don't use it in the exercise).

colinleach:
I'm currently getting lunch ready, but I'll make the changes this afternoon.

colinleach:
> with Task 1 and Task 4 swapping location

I tried that, but the tests for `shuffle_starships()` rely on task 2 to generate test values. I've swapped 3 and 4, as the best compromise I can think of.

exercises/concept/captains-log/.docs/instructions.md
colinleach: I've removed inclusive/exclusive from the instructions, adding a comment in the hints to not worry about it.

colinleach:
I _think_ I've made all the changes, but I admit I got a bit confused during all the changes. If you get chance, could you cast an eye over it and merge when you're happy?

depial:
> I tried that, but the tests for shuffle_starships() rely on task 2 to generate test values. I've swapped 3 and 4, as the best compromise I can think of.

The main thing I wanted to do was move Task 1 down, since it now has the most "complicated" answer for the tasks that use `sample`, but this is not a big deal.

> If you get chance, could you cast an eye over it and merge when you're happy?

Yep! I'll give it a closer look over tomorrow.

exercises/concept/captains-log/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+
+- This looks like [sampling][ref-sample] from an input vector.
+- Duplicates are possible, so sample with replacement.
+
+## 2. Randomize a fleet of starships
+
+- The Introduction discusses shuffling.
+
+## 3. Generate a random starship registry number
+
```
depial: These hints are out of order

exercises/concept/captains-log/.docs/instructions.md
```diff
@@ -0,0 +1,69 @@
...
+There is a full list of possible starships, but only a few of these is needed in each game, and they should be chosen at random.
+
+Implement the `shuffle_starships(starships)` function, which takes a vector of unique starship registry numbers, and returns a vector of the same length with starships in random order.
+It will then be easy for Mary to take a random selection of any size from the start of the vector.
+
+Result values should be taken from `starships`, with no duplicates.
+
+```R
+shuffle_starships(c("NCC-5011", "NCC-1228", "NCC-7039", "NCC-3978", "NCC-1476"))
+#> [1] "NCC-5011" "NCC-7039"
```
depial: This output should contain all the `starships`

depial:
I've flagged the two things that I found on a last look over.

One last question: Did I provide functionality in the test runner to support the "nested" test sets (i.e. the groups with more than one test)? I feel like I did, but I'm not sure anymore...

Edit: It appears [I did include that](https://github.com/exercism/r-test-runner/blob/main/bin/run.R#L46). Memory = no good => Comments in code = good :sweat_smile:

exercises/concept/captains-log/.docs/instructions.md
colinleach: It was on my to-do list. Unfortunately, I couldn't remember my to-do list...

colinleach:
I _think_ we've caught the problems, so I'll go ahead and merge. No doubt more problems will emerge later, but such is life.

PR #514, opened 2026-06-05
Update Captains log

depial:
The function name `random_planet` managed to slip through instead of `random_planet_class`.

Also, when completing the exercise online, I got a failure due to the statistics being too tight on Task 4, but it might have been a fluke. Not sure if we should change it.

### DIFF
--- a/exercises/concept/captains-log/test_captains-log.R
+++ b/exercises/concept/captains-log/test_captains-log.R
@@ -19,7 +19,7 @@ test_that("1. all generated planets are valid, with duplicates needed", {
   expect_all_true(random_planet_class(repeats) %in% planetary_classes)
 })
 
-test_that("1. most valid planets classes are generated", {
+test_that("1. planet classes are generated uniformly", {
   expect_all_true(planetary_classes %in% random_planet_class(repeats))
 })
 
--- a/exercises/concept/captains-log/test_captains-log.R
+++ b/exercises/concept/captains-log/test_captains-log.R
@@ -20,7 +20,7 @@ test_that("1. all generated planets are valid, with duplicates needed", {
 })
 
 test_that("1. most valid planets classes are generated", {
-  expect_true(sum(planetary_classes %in% random_planet_class(repeats)) >= 8)
+  expect_all_true(planetary_classes %in% random_planet_class(repeats))
 })
 
 # 2) random_ship_registry_number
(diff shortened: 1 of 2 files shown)


colinleach:
> Not sure if we should change it

I assume you mean your  Χ² stuff? It sounds bad if we're seeing a failure already, so I'm inclined to relax the stats. Probably best if you do it now?

colinleach (approved):

depial:
Yeah, the test statistics were calculated with 10_000 trials in Julia, which are now 1_000 in R, thereby invalidating the statistics. So, there are three options:

1. Change repeats back to 10_000
2. Recalculate the Χ² test statistics
3. Arbitrarily up the test statistics

I don't notice a huge lag on the web editor when running, so I'm inclined to try resetting the trials to 10_000 first. What are your thoughts?

colinleach:
Let's try 10k and see how well it works. The previous version of the tests was more complicated (and ran slower), so I backed off a lot on the repeats.

depial:
I updated the samples in Task 4 to 10_000, but left the other tasks at 1000 (where applicable). 

I also updated some of the other testing to better reflect some of the changes that we made. This includes both test names and how some are done. Please take a close look to see if they make sense.

One particular change was in the planetary classes. I've made the test a bit more stringent because there's only a 0.003% chance of only getting 9 out of the 10 planetary classes when sampling 1000 times.

exercises/concept/captains-log/test_captains-log.R
```diff
@@ -20,7 +20,7 @@ test_that("1. all generated planets are valid, with duplicates needed", {
 })
 
 test_that("1. most valid planets classes are generated", {
-  expect_true(sum(planetary_classes %in% random_planet_class(repeats)) >= 8)
+  expect_all_true(planetary_classes %in% random_planet_class(repeats))
```
colinleach, line 23: Maybe a bit _too_ stringent? I'd be more comfortable with `>= 9`. If you really want all, we'd better change the test string on line 22 to say "all", not "most".

colinleach:
I like most of the changes, just one remaining question. `p = 3e-5` for a false negative is good but not great.

depial:
> `p = 3e-5` for a false negative is good but not great.

Actually, I did the initial calculation wrong (entered 100 samples instead of 1000), and the p-value is actually closer to 1e-45 for finding only 9 out of 10 planets after 1000 samples.

colinleach:
OK, let's merge - over to you, when you're happy.

depial:
I've changed the wording of the test to `planet classes are generated uniformly` to better reflect what it's about.

depial:
There is noticeably more lag, but it appears to be fine, so I think we're good to go.

PR #515, opened 2026-06-10
add concept exercise trees-and-tribbleations

depial:
Here's a first draft of a concept exercise for `Dataframes`. Some notes:

- The functions `all_of` and `pick` are useful/needed to complete the exercise.
  - It appears `all_of` is necessary due to: `external vector in selections was deprecated`.
  - `pick` appears to be necessary since `arrange` didn't like strings.
  - I could see including the latter in the `about.md` and `introduction.md`, but maybe not the former?
- I've included a trimmed down `introduction.md`, but it's still pretty lengthy.
- Testing is currently a bit weak, but likely sufficient. I'm just worried that if people order their columns differently somehow, this could lead to a failure.

I had considered adding another column with "location numbers", so that people could practice with the `relocate(.after)` option, but I figured this might be over complicating things. I'm certainly open to suggestions on how to refine things or what else needs to be included.

Please pay attention to the docs, since I went through many variable name changes and I may not have caught them all. Also, let me know if the instructions are clear enough with a believable story, since I didn't spend a long time thinking about them.

### DIFF
--- a/exercises/concept/trees-and-tibbleations/.docs/hints.md
+++ b/exercises/concept/trees-and-tibbleations/.docs/hints.md
@@ -16,8 +16,7 @@
 
 - You need to move columns around and sort rows (see the Introduction).
 - The [`all_of()`][ref-allof] function may be useful for moving columns.
-- `arrange()` will silently fail if not given a proper column name (i.e. it doesn't like character strings)
-- `.data[[]]` or [`pick()`][ref-pick] might be useful for the sorting operation.
+- See caution in the introduction about [data-masking][ref-datamask] which mentions [`pick()`][ref-pick].
 
 ## 4. Customer copy of dataset
 
@@ -27,3 +26,4 @@
 
 [ref-allof]: https://tidyselect.r-lib.org/reference/all_of.html
 [ref-pick]: https://dplyr.tidyverse.org/reference/pick.html
+[ref-datamask]: https://rlang.r-lib.org/reference/args_data_masking.html
--- a/exercises/concept/trees-and-tibbleations/.docs/hints.md
+++ b/exercises/concept/trees-and-tibbleations/.docs/hints.md
@@ -15,11 +15,14 @@
 ## 3. Orchard copy of dataset
 
 - You need to move columns around and sort rows (see the Introduction).
-- The `all_of()` function may be useful for moving columns.
-- The `pick()` function might be useful for the sorting operation.
+- The [`all_of()`][ref-allof] function may be useful for moving columns.
+- The [`pick()`][ref-pick] function might be useful for the sorting operation.
 
 ## 4. Customer copy of dataset
 
 - You will need to take a subset of columns.
-- The `all_of()` function may be useful for subsetting.
+- The [`all_of()`][ref-allof] function may be useful for subsetting.
 - Only the trees between the minimum and maximum height *and* below the maximum weight should be in the copy.
+
+[ref-allof]: https://tidyselect.r-lib.org/reference/all_of.html
+[ref-pick]: https://dplyr.tidyverse.org/reference/pick.html
--- a/exercises/concept/trees-and-tibbleations/.docs/hints.md
+++ b/exercises/concept/trees-and-tibbleations/.docs/hints.md
@@ -1,0 +1,25 @@
+# Hints
+
+## 1. Rename a column
+
+- Remember, you need to modify the `trees` dataset, which is available if `library(datasets)` is called.
+- You may (or may not) want to convert the initial `data.frame` into a `tibble`.
+- See the Introduction for how to change a column name.
+
+## 2. Add girth and weight columns
+
+- See the Introduction for how to add columns.
+- The new columns can be calculated with existing columns.
+- Don't forget to round to the required precision.
+
+## 3. Orchard copy of dataset
+
+- You need to move columns around and sort rows (see the Introduction).
+- The `all_of()` function may be useful for moving columns.
+- The `pick()` function might be useful for the sorting operation.
+
+## 4. Customer copy of dataset
+
+- You will need to take a subset of columns.
+- The `all_of()` function may be useful for subsetting.
+- Only the trees between the minimum and maximum height *and* below the maximum weight should be in the copy.
--- a/exercises/concept/trees-and-tibbleations/.docs/hints.md
+++ b/exercises/concept/trees-and-tibbleations/.docs/hints.md
@@ -16,7 +16,8 @@
 
 - You need to move columns around and sort rows (see the Introduction).
 - The [`all_of()`][ref-allof] function may be useful for moving columns.
-- The [`pick()`][ref-pick] function might be useful for the sorting operation.
+- `arrange()` will silently fail if not given a proper column name (i.e. it doesn't like character strings)
+- `.data[[]]` or [`pick()`][ref-pick] might be useful for the sorting operation.
 
 ## 4. Customer copy of dataset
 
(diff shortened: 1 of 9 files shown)


depial:
Interesting that configlet is complaining about the lack of concepts now...

Edit: I just forgot to set the status to `wip`

colinleach:
I'm just heading out, so it may be a few hours before I can do a full review. Looking good so far!

exercises/concept/trees-and-tibbleations/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+## 2. Add girth and weight columns
+
+- See the Introduction for how to add columns.
+- The new columns can be calculated with existing columns.
+- Don't forget to round to the required precision.
+
+## 3. Orchard copy of dataset
+
+- You need to move columns around and sort rows (see the Introduction).
+- The `all_of()` function may be useful for moving columns.
```
colinleach: Should we add links to the docs for `all-of()` and `pick()`?

exercises/concept/trees-and-tibbleations/.docs/introduction.md
```diff
@@ -0,0 +1,304 @@
...
+  head(4)
+  # A tibble: 4 × 5
+#>   name           species height  mass   BMI
+#>   <chr>          <chr>    <int> <dbl> <dbl>
+#> 1 Luke Skywalker Human      172    77  26.0
+#> 2 C-3PO          Droid      167    75  26.9
+#> 3 R2-D2          Droid       96    32  34.7
+#> 4 Darth Vader    Human      202   136  33.3
+```
+
```
colinleach, line 281: I've added a section on `pick()` to #506 which I guess you'll want to copy to herre.

colinleach (approved):

colinleach:
I haven't found anything big, but we need to coordinate this with #506. Once you're happy with this Intro, I'll copy it to the concept.

depial:
I've taken quite a bit out of the exercise Intro, which might be more than necessary for the concept Intro. I'm not sure if you're going to want to copy it over directly or maybe try to whittle down the `about.md` in a similar way.

colinleach:
I'm not sure whether we need to include `library(datasets)`. On both my Linux and Windows installations, it seems to be pre-loaded. Anyway, it does no harm, and we can experiment in the online editor once it's merged.

exercises/concept/trees-and-tibbleations/.docs/introduction.md
```diff
@@ -0,0 +1,322 @@
...
+#>   <chr>          <chr>    <int> <dbl> <dbl>
+#> 1 Luke Skywalker Human      172    77  26.0
+#> 2 C-3PO          Droid      167    75  26.9
+#> 3 R2-D2          Droid       96    32  34.7
+#> 4 Darth Vader    Human      202   136  33.3
+```
+When you want to operate on a subset of the columns with functions such as `mutate()`, the `select() |> mutate()` sequence in the above example is one option.
+Only the selected columns will be in the result.
+
+Alternatively, it can be convenient to use [`pick()`][ref-pick] _within_ the `mutate()` call:
```
colinleach: Link is now broken

colinleach:
I just found a broken link in the Intro. I guess we just remove it.

depial:
I've added more hints for Task 3 in regards to `arrange` not liking strings. I think technically we should use `.data[[]]` to convert the string, but `pick()` works likewise. 

Could you have a look at those hints to see if they are sufficient or need more?

depial:
I'm starting to wonder if we should include information in the `about.md` and `introduction.md` on the behavior of [`arrange()` with strings](https://stackoverflow.com/questions/27034655/how-to-use-dplyrarrangedesc-when-using-a-string-as-column-name)?

I'm not sure what the canonical way to do it is, but there is a comment at the end of that stack overflow post which points to `pick()` as being the new documented way.

colinleach:
This is getting complicated enough that we probably need a short note in the Intro, not just hints, to say that pick is more flexible. I'd need to think more about _what_ to say.

depial:
Not sure if it's any help (or you already know about all this), but I tried to use `pick()` by itself, and it said it has to be used in "data-masking verbs like mutate()", while `arrange()` expects "<[data-masking](https://rlang.r-lib.org/reference/args_data_masking.html)> Variables, or functions of variables" as its second argument.

I can get a feeling for what that might mean, but I'm assuming it's far too technical for a short note in the Intro.

colinleach:
I tend to feel an Intro shouldn't contain things I don't understand myself!

colinleach:
With what we now know, do we still want to get into all this in a Concept exercise? Maybe yes, but it's becoming tempting to just have a simpler exercise where the instructions specify columns to move, instead of passing them as an argument.

Pity, it seemed such a good idea...

depial:
I'm actually starting to feel like it's a good opportunity to make people aware of this issue. I nearly missed it since `arrange` fails silently and it seems like this could be a fairly common mistake.

I think if we add a line to the `arrange` section to the effect of:

> Many functions, such as `arrange()` are data-masking and require [data-masking variables](https://rlang.r-lib.org/reference/args_data_masking.html).
For this reason, non-data-masking input (e.g. character strings) needs to be converted to be used in data-masking functions.
A full treatment of data-masking is beyond the scope of this concept, but it's useful to know there are ways of making this conversion which include `pick()`, `.data[[]]` and `!!sym()`.

We could probably put that in an `exercism/caution` block.

I'm not sure how you feel about that write up, but I figure between that and some hints, we should be okay.

Let me know if you're still feeling uneasy about it. I'm not totally against specifying the columns in the instructions, since that's probably closer to how people write scripts in the wild anyway, but it's just strikes me as odd in that case, since a function is then only necessary for us to make testing modular.

colinleach:
OK, let's go with the extra documentation and the original exercise.

depial:
I've added the language to the end of the `introduction.md` along with some examples showing the silent error and the correct usage of `pick()`. I've also gone ahead and removed the `mutate()` example, since the `arrange()` one should better reflect what's needed in the exercise.

I've also added a link to the data-masking variables page in the hints and referred back to the introduction. Let me know if you think that's good enough or we need more verbose hints.

exercises/concept/trees-and-tibbleations/.docs/introduction.md
```diff
@@ -0,0 +1,330 @@
...
+#> 1 Fortran      1957 FALSE       
+#> 2 Julia        2012 TRUE        
+#> 3 Python       1991 TRUE        
+#> 4 R            1993 TRUE     
+```
+
+~~~~exercism/caution
+Many functions, such as `arrange()` and `mutate()` are data-masking and require data-masking variables.
+For this reason, non-data-masking arguments (e.g. character vectors) need to be converted to be used in data-masking functions.
+A full treatment of how data-masking works in R is beyond the scope of this concept, but it's useful to know there are ways of making this conversion which include options such as: `pick()`, `.data[[]]` and `!!sym()`.
```
colinleach, line 310: Maybe add a blank line before 309 to break up the text?

exercises/concept/trees-and-tibbleations/.docs/introduction.md
```diff
@@ -0,0 +1,330 @@
...
+~~~~
+
+```R
```
colinleach, line 312: Does the example belongs inside the caution block? I could argue it either way, and we'll get a better idea when we see how the website displays it.

depial: I agree it belongs in the caution block, so I'll include it there and if it looks bad, we can take it out later.

colinleach:
I think it's good to go. Merge when you're happy.

I'll update the concept, though I'm not sure if I'll get to it today (I woke up tired, and the trend through the day is predictable). I'm making good progress with `cater-waiter` in the last hour, so I'd like to finish that before I forget what I was doing.

depial:
> I'm making good progress with cater-waiter in the last hour, so I'd like to finish that before I forget what I was doing.

All the best to you there! I'm calling it a day here.

PR #501, opened 2026-04-29
[New Concept] Functional Programming

colinleach:
In #493, I implied that a PR of Functional Programming was imminent, and we should review them as a pair. A mere 6 weeks later, here it is.

Concept exercises are not an obvious choice.
- For Functions, `locomotive-engineer` is probably not a great choice for R, but I don't know what is.
- For Functional Programming, we could probably modify `julia/cheese-club` to fit.

Until we know the exercises, I've left the Introductions blank.

Also, my visitors arrive 2 days from now, staying till next Wednesday, so I'll be otherwise occupied.

PR #518, opened 2026-06-12
coordinate-transformation: possible functions exercise?

colinleach:
This is very incomplete, but might it be the starting point for an exercise? I'm OK with junking it if we think of something better. I'd like to keep it quite short and simple, though. This is a bitty concept of various not-very-related things student need to know about, so reading the docs is the main point.

The Functions concept #501 is relatively early in the syllabus compared with stuff we've worked on recently (right after `vector-functions`), so no lists or switch or loops allowed.

Don't pay too much attention to the Introduction. I've added an extra bit on functions as return values, but haven't yet taken anything out - so it's longer than the concept About at this stage.

The sample code is still in `coordinate-transformation.R`, not yet moved to `.meta/exemplar.R`.

### DIFF
--- a/exercises/concept/coordinate-transformation/.docs/hints.md
+++ b/exercises/concept/coordinate-transformation/.docs/hints.md
@@ -1,0 +1,3 @@
+# Hints
+
+TODO
--- a/exercises/concept/coordinate-transformation/.docs/hints.md
+++ b/exercises/concept/coordinate-transformation/.docs/hints.md
@@ -1,3 +1,26 @@
 # Hints
 
-TODO
+## 1. Scale the coordinates
+
+- The `point` is a vector, scaling is by a vector `s` (either explicitly a vector, or by recycling). - You already know about element-wise multiplication of vectors.
+
+## 2. Translate the coordinates
+
+- Convert the dot args to a vector.
+- The usual [`c()`][ref-c] function is flexible enough to handle this.
+- Unpacking vector elements into individual variables is typically _not_ a good way to use R: think in terms of vector operations.
+
+## 3. Transform a 2D point
+
+- Remember: translate, _then_ scale.
+- Arguments to `transform2d()` are optional, so remember to define defaults.
+- `transform2d()` needs to return a function which takes a `point` as its sole argument.
+- If still confused, read the Introduction on environments and functions as return values, and look at the examples in the instructions.
+- This is the sort of topic that can confusing at first, but becomes very natural with a bit of practice.
+
+## 4. Transform a 3D point
+
+- This is just an extension of task 3.
+- It relies on tasks 1 and 2 being implemented in a simple, flexible way.
+
+[ref-c]: https://www.rdocumentation.org/packages/base/versions/3.3.0/topics/c
--- a/exercises/concept/coordinate-transformation/.docs/instructions.md
+++ b/exercises/concept/coordinate-transformation/.docs/instructions.md
@@ -25,7 +25,7 @@ Implement the `translate(point, ...)` function that returns a new point, with ea
 point <- c(2, 3)
 
 # supply dx, dy translations
-translate(point, .5, .6)
+translate(point, 0.5, 0.6)
 #> [1] 2.5 3.6
 ```
 
@@ -34,6 +34,8 @@ translate(point, .5, .6)
 Some of your teammates are less experienced with R, so you decide to use a function closure to create reusable transformations for `{x, y}` coordinate pairs.
 
 Implement the `transform2d(dx, dy, s)` function that returns a function making use of a closure to perform a repeatable 2d translation and scaling of a point.
+
+All arguments should be optional, with default values of `0.1` for `dx` and `dy`, and `1` for `s`.
 
 The returned function should take a 2D point, then:
 
@@ -44,12 +46,12 @@ The returned function should take a 2D point, then:
 # scale all coordinates by the same amount (2) in this example
 # your code needs to be able to handle a vector `s`
 
-f <- transform2d(.5, .6, 2)
+f <- transform2d(0.5, 0.6, 2)
 class(f)
 #> [1] "function"
 
-f(2, 3)
-#> [1] 5 7.2
+f(c(2, 3))
+#> [1] 5.0 7.2
 ```
 
 The order of operations is important: translate-then-scale gives the correct result, scale-then-translate does not.
@@ -61,10 +63,13 @@ Mapping data can also contain heights, used to add shading and contours on scree
 
 Implement the `transform3d(dx, dy, dz, s)` function that returns a function.
 
-The returned function should take a 3D point, and translate it by the pre-defined values, then scale it by `s`.
+All arguments should be optional, with default values of `0.1` for `dx`, `dy` and `dz`, and `1` for `s`.
+
+The returned function should take a 3D point, translate it by the pre-defined values, then scale it by `s`.
 
 ```R
-g <- transform3d(.5, .6, 0.7, 2)
-g(c(2, 3, 4))
-#> [1] 5 7.2 9.4
+g <- transform3d(0.5, 0.6, 0.7, 2)
+point <- c(2, 3, 4)
+g(point)
+#> [1] 5.0 7.2 9.4
 ```
--- a/exercises/concept/coordinate-transformation/.docs/instructions.md
+++ b/exercises/concept/coordinate-transformation/.docs/instructions.md
@@ -1,67 +1,70 @@
 # Instructions
 
-Your design company has primarily been working with CSS transformations to build web pages.
+Your company has been working with mapping data to build web pages.
 After some discussion, a decision is made to start using R to perform some calculations dynamically.
 
-Some of your teammates are less experienced with R, so you decide to use a function closure to create reusable transformations for `{x, y}` coordinate pairs.
+Please aim to work directly with the vector points: there is no need to unpack them into separate `x` and `y` coordinates.
 
-## 1. Move the coordinates
+## 1. Scale the coordinates
 
-Implement the `move(x, y, dx, dy)` function that returns a new point `c(x1, y1)`, with `x, y` moved by `dx, dy`.
-
-Arguments `x` and `y` are rquired.
-`dx` and `dy` are optional, with both defaulting to `0.1`
+Implement the `scale(scaling)` function that takes a point with arbitrary number of dimensions, and scales it by the pre-defined `scaling`.
 
 ```R
-move(2, 3, .5, .6)
-#> [1] 2.5 3.6
-
-move(2, 3)
-#> [1] 2.1 3.1
+# A 2D example
+scaling <- c(0.5, 0.8)
+point <- c(10, 5)
+scale(point, scaling)
+#> [1] 5 4
 ```
 
 ## 2. Translate the coordinates
 
-Implement the `translate2d(dx, dy)` function that returns a function making use of a closure to perform a repeatable 2d translation of a coordinate pair.
-
-The returned function should take `x` and `y` coordinates of a point, and translate it by the pre-defined `dx` and `dy`.
+Implement the `translate(point, ...)` function that returns a new point, with each coordinate moved by the values in the ellipses.
 
 ```R
-f <- translate2d(.5, .6)
+point <- c(2, 3)
+
+# supply dx, dy translations
+translate(point, .5, .6)
+#> [1] 2.5 3.6
+```
+
+## 3. Transform a 2D point
+
+Some of your teammates are less experienced with R, so you decide to use a function closure to create reusable transformations for `{x, y}` coordinate pairs.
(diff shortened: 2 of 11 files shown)


exercises/concept/coordinate-transformation/.docs/introduction.md
```diff
@@ -0,0 +1,247 @@
...
+Below is a preview, using [`sapply()`][ref-sapply] to square each number in a range:
+
+```R
+sapply(1:5, \(x) x ^ 2)
+#> [1]  1  4  9 16 25
+```
+
+That is not a very useful example, because `(1:5) ^ 2` returns the same result more simply, but it illustrates how we define a function without bothering to think of a name for it.
+
+## Functions as return values
```
depial, line 129: I feel like this section should abut `Function Environment`. I could see the following options for arrangements:

- `Anonymous Functions` -> `Function Environment` -> `Functions as return values`
- `Anonymous Functions` -> `Functions as return values` -> `Function Environment`

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,67 @@
+# Instructions
+
+Your design company has primarily been working with CSS transformations to build web pages.
```
depial: The context could probably be changed to fit better with R. Maybe something like:

> Your company has been working with mapping data...

or something else that has a data science / stats feel to it.

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,67 @@
...
+
+Implement the `translate_point(delta)` function that returns a function .
+
+The returned function should take a 2D point, and translate it by the pre-defined `delta`.
+
+```R
+g <- translate_point(c(.5, .6))
+g(c(2, 3))
+#> [1] 2.5 3.6
+```
```
depial: I feel like this task is a bit repetitive. Maybe we could find an exercise where the ellipsis can be used to replace it?

exercises/concept/coordinate-transformation/.docs/introduction.md
colinleach: We need the order Function Environment -> Functions as return values, because the return value captures the environment.

exercises/concept/coordinate-transformation/.docs/instructions.md
colinleach: Yeah, this was pretty much a TODO with a Clojure document.

depial:
I think this could work. I've added a couple of comments, but the other thing I'm unsure about is having three tasks with closures. Unless we really want to focus on those, I would probably write this as a couple of normal functions (translation, scaling, etc) and then the last function could be a wrapper for these that is a closure. Something to the effect:

```R
transform <- function(dx, dy, sx, sy) {
  \(point) scale2d(translate2d(point, dx, dy), sx, sy)
}
```

I could also imagine using only vector input for the points. That way we could potentially include an ellipsis for general `translate` and `scale` functions:

```R
translate <- function(point, ...) {
  point + c(...)
}

translate(c(1,2), 2, 3) # 2D
#> [1] 3  5

translate(c(1,2,3), 2, 3, 4) # 3D
#> [1] 3  5  7
```

This would make the `transform` function a bit more difficult, but we could have two forms of that: `transform2d` and `transform3d` which reuse the two more general `translate` and `scale` functions. This brings us back to four tasks with two of them being closures.

<details>
<summary>possible exemplar.R</summary>

```R
scale <- function(point, s) {
  point * s
}

translate <- function(point, ...) {
  point + c(...)
}

transform2d <- function(dx=0.1, dy=0.1, s=1) {
  \(point) scale(translate(point, dx, dy), s)
}


transform3d <- function(dx=0.1, dy=0.1, dz=0.1, s=1) {
  \(point) scale(translate(point, dx, dy, dz), s)
}
```
</details>

Let me know what you think.

exercises/concept/coordinate-transformation/.docs/instructions.md
colinleach: 2, 3, 4 are repetitive. I'd like to keep some element of going to vector input rather than individual arguments: this is quite early in the syllabus, so vector arithmetic is still worth a bit of practice.

Getting an ellipsis into it seems good in principle. I need more caffeine before thinking about the details...

colinleach:
> Let me know what you think.

Coffee first, thought later

depial:
I think I've got to do the same, since I've messed up the ideas for how scaling and translating work... So I'll have some tea and try to correct that :D

Edit: Okay, the math should make more sense in my example above now

exercises/concept/coordinate-transformation/.docs/introduction.md
depial: It's just `Anonymous Functions` which is separating the two, so I'd suppose another possible ordering could be:

`Function Environment` -> `Functions as return values` -> `Anonymous Functions`

colinleach:
Not the most productive morning (the coffee needed Tylenol backup), but this is where I'm up to. Still plenty of rough edges, but I think it's much improved.

depial:
I think it's looking good. 

The one thing I'd like to get into the exercise is a default argument. It only really makes sense to do that in the wrapper functions, so maybe we could request `s = 1` for a default in Tasks 3 and 4? That way they would be translation functions by default.

colinleach:
> The one thing I'd like to get into the exercise is a default argument

I tested for it, I just forgot to mention it in the instructions. Like I said, not the most productive morning.

depial:
> Like I said, not the most productive morning.

Understandable! I hope you're feeling better!

> I tested for it

After thinking about it a bit, I kind of like the idea of only having a default for the scaling (since it has the extra explanation of making the transform function a translation function by default). That said, I don't really care at all, so I'm happy to go whichever way you'd like to.

colinleach:
I made the latest set of changes before reading your last comment. See what you think, we can change it as necessary.

Also, after marking this as ready for review I remember that I intended to beef up the (currently very sparse) tests. Let's decide on the default-argument stuff first, then I'll do the tests.

exercises/concept/coordinate-transformation/.docs/hints.md
```diff
@@ -0,0 +1,26 @@
...
+## 2. Translate the coordinates
+
+- Convert the dot args to a vector.
+- The usual [`c()`][ref-c] function is flexible enough to handle this.
+- Unpacking vector elements into individual variables is typically _not_ a good way to use R: think in terms of vector operations.
+
+## 3. Transform a 2D point
+
+- Remember: translate, _then_ scale.
+- Arguments to `transform2d()` are optional, so remember to define defaults.
```
depial: This may need rewording if we switch to only `s` being optional.

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
+# Instructions
+
+Your company has been working with mapping data to build web pages.
+After some discussion, a decision is made to start using R to perform some calculations dynamically.
+
+Please aim to work directly with the vector points: there is no need to unpack them into separate `x` and `y` coordinates.
+
+## 1. Scale the coordinates
+
+Implement the `scale(scaling)` function that takes a point with arbitrary number of dimensions, and scales it by the pre-defined `scaling`.
```
depial: I think this signature needs to be:

> Implement the `scale(point, scaling)` function that takes

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
...
+
+Implement the `translate(point, ...)` function that returns a new point, with each coordinate moved by the values in the ellipses.
+
+```R
+point <- c(2, 3)
+
+# supply dx, dy translations
+translate(point, 0.5, 0.6)
+#> [1] 2.5 3.6
+```
```
depial, line 38: I think it would be instructive to put an example with three dimensions:

```R
point2d <- c(2, 3)

# supply dx, dy translations
translate(point2d, 0.5, 0.6)
#> [1] 2.5 3.6

point3d <- c(2, 3, 4)

# supply dx, dy, dz translations
translate(point3d, 0.5, 0.6, 0.7)
#> [1] 2.5 3.6 4.7
```

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
...
+# A 2D example
+scaling <- c(0.5, 0.8)
+point <- c(10, 5)
+scale(point, scaling)
+#> [1] 5 4
+```
+
+## 2. Translate the coordinates
+
+Implement the `translate(point, ...)` function that returns a new point, with each coordinate moved by the values in the ellipses.
```
depial, line 22: It might not hurt to add a line explaining that this function should take points of any dimension.

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
...
+#> [1] 2.5 3.6
+```
+
+## 3. Transform a 2D point
+
+Some of your teammates are less experienced with R, so you decide to use a function closure to create reusable transformations for `{x, y}` coordinate pairs.
+
+Implement the `transform2d(dx, dy, s)` function that returns a function making use of a closure to perform a repeatable 2d translation and scaling of a point.
+
+All arguments should be optional, with default values of `0.1` for `dx` and `dy`, and `1` for `s`.
```
depial: Also, would need rewording on an optional argument change.

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
...
+```R
+# scale all coordinates by the same amount (2) in this example
+# your code needs to be able to handle a vector `s`
+
+f <- transform2d(0.5, 0.6, 2)
+class(f)
+#> [1] "function"
+
+f(c(2, 3))
+#> [1] 5.0 7.2
```
depial: Maybe we could change the function `f` to something like `tf2d`, `trsfrm2d` or something like that?

Also, it would probably be good to show an example with an optional argument:

```R
tf2d <- transform2d(0.5, 0.6)

tf2d(c(2, 3))
#> 2.5 3.6
```

exercises/concept/coordinate-transformation/.docs/instructions.md
```diff
@@ -0,0 +1,75 @@
...
+
+All arguments should be optional, with default values of `0.1` for `dx`, `dy` and `dz`, and `1` for `s`.
+
+The returned function should take a 3D point, translate it by the pre-defined values, then scale it by `s`.
+
+```R
+g <- transform3d(0.5, 0.6, 0.7, 2)
+point <- c(2, 3, 4)
+g(point)
+#> [1] 5.0 7.2 9.4
```
depial: Maybe `g` to `tf3d` or `trsfrm3d`?

Also, if you wanted, you could show an example with a vector `s` (e.g. `s = c(1,2,3)`).

exercises/concept/coordinate-transformation/.docs/introduction.md
```diff
@@ -0,0 +1,209 @@
...
+f(4)
+#> [1] 12
+```
+
+## Anonymous Functions
+
+When we define a function, we usually bind the resulting function object to a variable:
+
+```R
+squareit_short <- function(x) x ^ 2
```
depial: This could make more out of the previous example by doing:

```R
x <- 2

doubleit_short <- function(y) x * y
```

And it could be mentioned that this is effectively doing the same thing as the previous example, but using a constant instead of a variable (or something to that effect).

depial (approved):
This looks good!

I'm leaning more towards the single optional argument with `s = 1`, just because it seems easier to understand for me. Again, I'm not wedded to the idea, so if you have a preference, feel free to express it.

exercises/concept/coordinate-transformation/.docs/instructions.md
depial: Also, maybe `scaling` should be changed to `s` (or vice versa) to be consistent with the rest of the exercise

colinleach:
All good suggestions. However, I'm quite likely to mess up if I make the changes now, so tomorrow is probably more appropriate.

exercises/concept/coordinate-transformation/.docs/introduction.md
colinleach: I changed the example, with a comment about capturing `x` from the environment. I tried to add the lnk back to the previous example, but it looked weird and I gave up.

colinleach:
I made a small change to the Intro: some but not all of what you suggested. If we get to agreement on this sort of change, I'll add it back to the concept About and Intro. Might as well be in this PR, rather than starting a separate one.

depial:
Looks great! You can merge when you're ready

PR #523, opened 2026-06-14
more links in json list

colinleach:
Just scratching the surface of what's needed, but it might give the Exercism website a prod to include this in the syllabus.


depial (approved):

depial:
I thought that would do the trick, but it doesn't appear to have worked...

colinleach:
Bummer

PR #525, opened 2026-06-15
delete unwanted links from hints

colinleach:
A trivial bit of cleanup: with data masking removed from the exercise we no longer need the references.

Mostly, I'm still trying to prod the website to add this to the syllabus.


depial (approved):

colinleach:
Looking good now!

PR #526, opened 2026-06-15
Update lasagna introduction.md

depial:
I've brought in more of the `about.md` since it seems to be more explanatory than what made it into the exercise `introduction.md`. I think this should address most of the concerns of @BNAndras.

If this seems alright, I can copy this over to the concept `introduction.md` as well.

### DIFF
--- a/concepts/basics/about.md
+++ b/concepts/basics/about.md
@@ -19,7 +19,7 @@ class(50L)
 #> [1] "integer"
 ```
 
-~~~exercism/advanced
+~~~~exercism/advanced
 You may also see references to `mode()`, which is similar to `typeof()`.
 
 Please ignore `mode()` and `storage.mode()`.
--- a/concepts/basics/introduction.md
+++ b/concepts/basics/introduction.md
@@ -10,7 +10,7 @@ avogadro <- 6.02e23
 mystr <- "Hello, World!"
 ```
 
-The preferred assignment operator is `<-` (not `=`).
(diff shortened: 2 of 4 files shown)


depial:
We could also include a [link](https://stat.ethz.ch/R-manual/R-devel/library/base/html/assignOps.html) to the docs on the matter in the `about.md`.

colinleach (approved):

colinleach:
> I can copy this over to the concept introduction.md as well.

Yes, please.

> We could also include a [link](https://stat.ethz.ch/R-manual/R-devel/library/base/html/assignOps.html) to the docs

My first thought was yes, we should do it. Then I looked at the page we're linking. A bit _heavy_ for a Basics Concept, perhaps? `<<-` is for assignment to global variables, which I'm not keen to encourage. Right assign just seems odd, and I don't remember ever seeing it used.

Some things "seemed like a good idea at the time", and have to be kept around for compatibility.

depial:
I'm fine with leaving the link out since I agree the description is pretty dense.

I think `<<-` might be a bit more than just global variable assignment, since it seems to be "scope searching" too (for lack of a better term). For example, the times I've found it useful/necessary is for [side effects outside of a scope](https://github.com/exercism/r-test-runner/blob/main/bin/run.R#L56). Here, the assignment is being made to something at the top-level scope (on [line 22](https://github.com/exercism/r-test-runner/blob/main/bin/run.R#L22)), and doesn't work otherwise.

In any case, I don't think we need to mention it here, so I'll go ahead and merge.

PR #532, opened 2026-07-12
Bump workflows to Ubuntu 26.04 and r-version 4.6.1

IsaacG:


colinleach (approved):

colinleach:
I have no immediate suggestions for why this is failing. I'll look into it, but suggestions welcome.

colinleach:
The problem is within the built-in action `r-lib/actions/setup-r-dependencies`. I have no idea how to fix whatever GitHub is doing.

How important is the Ubuntu 24.04 to 26.04 at this time? The older version is still LTS until May 2029, and it works. R installations have a lot of (rather brittle) moving parts, and we mess with them at our peril.

I'm tempted to PR an R upgrade to 4.6.1, without changing the OS. Partly out of curiosity to see if it errors.

BethanyG:
Looks like it can't find `pak` (was that phased out for 26.04?:

 https://github.com/exercism/r/actions/runs/29698236915/job/88222564487?pr=532
https://github.com/exercism/r/actions/runs/29698236913/job/88222564567?pr=532

colinleach:
> Looks like it can't find pak

I don't _think_ so. The pak GH repo has been updated within the last 3 days. In any case, we would need GH to fix their built-in action.

colinleach:
#533 makes the R change without the Ubuntu change. CI is still running, but it already got a lot further than this #532. This reinforces my gut feeling (uneducated prejudice) that an early upgrade to Ubuntu 26.04 is more trouble than it's worth.

Maybe try again in a couple of months, and see what changes?

_Update:_ #533 completed successfully, Bethany approved, we can merge.

BethanyG:
Shall we close this then?

colinleach:
> Shall we close this then?

I'm in favor, is @IsaacG?

IsaacG:
> > Shall we close this then?
> 
> I'm in favor, is @IsaacG?

You and Bethany are R maintainers. I'm not. I defer to y'all!

If Ubuntu 26 "just works", it's nice to bump it. If there's incompatibilities, there's no rush to bump it. Eventually we'll need to update it. If something doesn't work on 26, we can hope upstream fixes the issue or we'll need to figure it out at some point, so we probably don't want to forget about this until we realize 24 is no longer supported. But if it helps to wait 6 months, then there's no rush!

BethanyG:
Probably for another day, but if the runner uses `r-minimal`, perhaps we should use it here as well.  Can't remember off the top of my head, but I think on the Python runner we do this thing where we use Ubuntu (runs on), but then pull the Alpine container to do the actual testing (using container:  ____), plus the `run-in-docker.sh` script.
https://github.com/exercism/python-test-runner/blob/main/.github/workflows/main.yml#L43. 

Perhaps we can look at doing that here as well.  Or just wait for GH to un-munge their script.  Either way, `pak` is essential, since it's the R package manager du jour.

IsaacG:
There are a couple of tracks that pull in the Exercism Docker image for testing. It certainly makes setup easier! 

There's at least one track that explicitly runs tests on both the Exercism Docker image and the GHA image. It's possible the intent here is to test a wider range of environments as students might be running the tests locally in different settings. 

Either approach makes sense, depending on what exactly you want to test.

PR #539, opened 2026-08-05
[Update exercise] Crypto Square

colinleach:
I solved this exercise in R yesterday, in preparation for adding it to Julia and Fortran. Turns out, the 8-year-old implementation diverges a lot from canonical data: over-complicated, restrictive and really quite annoying.

I've added a template, updated tests to the current problem specification, and removed all functions except `ciphertext()`.

This should give students more freedom to explore different approaches. I don't _think_ it breaks existing solutions, it just ignores big chunks of code.

### DIFF
--- a/exercises/practice/crypto-square/.meta/example.R
+++ b/exercises/practice/crypto-square/.meta/example.R
@@ -11,7 +11,7 @@ ciphertext <- function(input) {
   normalized |> 
     str_pad(c * r, "right") |>
     str_split_1("") |>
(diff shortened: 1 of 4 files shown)


colinleach:
There are 108 completions, so retesting should be reasonable.

depial (approved):

depial:
A bit odd that it erred with `tidyverse`. Any thoughts on that?

colinleach:
My fault: sloppy copy-paste.

PR #541, opened 2026-08-27
[New Concept] Classes

colinleach:
I started this months ago, so I need to get it ticked off as something I no longer need to worry about. It was certainly educational, in the sense that my knowledge of the concept increased slightly, from zero to not-zero.

I haven't thought about an exercise yet. For now, I'll be content to get an agreed `about.md` then copy that (in full) to the `introduction.md`.

We'll eventually want to link `robot-simulator` as a practice exercise. The [implementation](https://github.com/exercism/r/blob/main/exercises/practice/robot-simulator/.meta/example.R) of that may give us ideas for a concept exercise: change the story (toy car?), simplify, break it down into tasks, etc.


concepts/classes/.meta/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "authors": [
+    "colinleach"
+  ],
+  "contributors": [],
+  "blurb": "R has several types of OOP class, but S3 is the simplest and most widely used."
```
depial: I was initially a bit confused by the use of `OOP class` being used in the singular since I don't think I would use it in this categorical/uncountable way. I looked into it, and it's technically correct, while also potentially being more common in British English than American English.

Feel free to leave it as is, especially if it is idiomatic, but I'll offer one suggestion which sounds more natural to my (naive) American ear:

> R has several types of OOP systems, but S3 is the simplest and most widely used.

Or, another which is more comfortable to my ear is your phrasing from the `about.md` which has, "...multiple class designs...".

concepts/classes/about.md
```diff
@@ -0,0 +1,212 @@
...
+By this point in the syllabus, you have seen R used often as an imperative language, and increasingly as a functional language (especially with the [`purrr`][ref-purrr] and [`dplyr`][ref-dplyr] libraries).
+
+Suggesting now that R is primarily object-oriented may provoke some skepticism!
+
+Oddly, this is a true statement.
+Internally, R is implemented using classes, inheritance, polymorphism, and the sort of attitudes typical of OOP programmers.
+
+However, anyone familiar with Java or Ruby (explicitly OOP) or even Python (semi-covertly OOP, in some respects) will find the R approach ... _unfamiliar_.
+
+For a start, there are multiple class designs: perhaps three in common use (S3, S4, R6), plus various others of mainly theoretical interest.
```
depial, line 12: Reference to my comment on the blurb

concepts/classes/about.md
```diff
@@ -0,0 +1,212 @@
...
+#> [1] "point2d"
+attributes(pt2d)
+#> $names
+#> [1] "x" "y"
+
+#> $class
+#> [1] "point2d"
+```
+
+So [`class()`][ref-class] just adds one attribute, in this case identifying `pt2d` as an instance of the `point2d` class.
```
depial: This is kinda interesting how `class()` appears to be performing double duty:

1. Return the class
2. Assign the class

This is certainly shown in the example, but it might be nice to explicitly mention it in this phrase. If nothing else, we may want to specify a bit more with something like:

> So making an assignment to [`class(obj)`][ref-class] just adds one attribute to `obj`, in this case identifying `pt2d` as an instance of the `point2d` class.

concepts/classes/about.md
```diff
@@ -0,0 +1,212 @@
...
+```R
+print.point2d <- function(pt) {
+  cat("Cartesian coords (", pt$x, ", ", pt$y, ")\n", sep="")
+}
+
+pt2d
+#> Cartesian coords (2.5, 4.2)
+```
+
+Now suppose we want to create a new generic method, for example `flipvert()` to invert the signe on the `y` coordinate.
```
depial: Typo in `signe`

concepts/classes/about.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Inheritance
+
+OOP enthusiasts can write multiple long textbook chapters on inheritance, subclassing, mixins, and all the rest.
+
+R just lets you supply multiple names to the `class` attribute.
+
+At the risk of stretching the example to breaking point, suppose we had made a generic `point` class, and wanted to subclass it to `point2d`, `point3d`, etc.
+
+Then the `point2d` definition simply uses `class = c("point2d", "point")`: the specific class and the parent class (or classes).
```
depial, line 186: Connected to my comment on assignment from above, one question that came to mind: I've now seen that we can assign strings and vectors of strings for the class, is there anything else?

I suspect there isn't, so we could probably make this explicit or not without much difference.

depial (approved):
Looks good! I found it quite informative for my level of current knowledge of the class system in R (which was near-zero).

I've included a couple of possible suggestions, and for some reason I was feeling quite verbose today, so try not to feel as though they are more imperative simply because there's more content; other than the typo, everything is a soft suggestion.

concepts/classes/.meta/config.json
colinleach: Hastily retrofitting an excuse: New Zealanders in general, and the University of Auckland in particular, have  an outsize influence on R development, so the language is designed with UK/Kiwi spelling and US spelling supported interchangeably. Though this is not Exercism policy.

concepts/classes/about.md
colinleach: I've added an extra few words about vectors of string (I should really call them `character` vectors, but I hate that terminology).

PR #125, opened 2019-08-06
Affin cipher fix

tintinthong:
This is my fix to the previous pull-request on #123. 

Using GitKraken as GUI tool for Git, created new branch `affin-cipher-fix`.  The problem was that my previous branch `affine-cipher` had excessive historical commits that was squashed on the upstream remote master(exercism/r/master).  It is a mouthful and confusing but I basically, reset to the commit 73af394, pulled canges from upstream remote master and cherry picked my pull request commits 0ba2c4 and 950598 on `affine-cipher` branch. Hope I didn't change any history that has any conflicts.

In this new commit, I made code more readable by breaking it up into lines (also because lintr bot also said my line had too many characters).

Once everything is fine. I will do a squash myself.

PR #545, opened 2026-08-29
Affine cipher

colinleach:
Replaces #125, which addresses a much earlier version of this exercise.

In the end, not much of the original code survived to the working solution.

Necessary changes:
- Numerical digits are now passed through unchanged.
- Output from `encode()` is in 5-character chunks.

Optional changes:
- Functional design with `stringr`, `purrr` and lots of pipes.
- Refactoring to reduce code duplication.

I'm undecided between difficulty 4 or 5. This was discussed (without resolution?) in #125, and other tracks vary widely. Whatever, I think it's somewhere in the Medium range.

I'll also create a Julia version, while I remember where the various gotchas are.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1107,6 +1107,14 @@
         "difficulty": 4
       },
       {
+        "slug": "affine-cipher",
+        "name": "Affine Cipher",
+        "uuid": "478b838b-dc33-4532-984c-281c0c2424b8",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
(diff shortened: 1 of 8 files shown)


depial (approved):
LG2M!

depial:
I'm also not sure about the difficulty. i saw the original conversation, and I could go either way. I guess people will feel differently depending on their grasp of mathematics, but I'm fine leaving it as a 5 either way.

PR #548, opened 2026-09-02
[Line Up] sync tests

colinleach:
There are a bunch of new tests in `canonical-data`. This updates the exercise, same as we just did on the Julia track.


colinleach:
21 completions on R, exactly the same as Julia. Cheap to retest.

depial (approved):

